PR #28, opened 2016-05-31
Update the main text color and update info-lite brand color

i386:
* Body and table header colour is `brand-grey-lite` 
* Updated the `brand-info-lite`, `brand-grey` and `brand-grey-lite` to match the official palette
* Buttons should use regular rather than semibold to avoid smudging since apparently the font is wrong ;)

@reviewbybees


sophistifunk:
🐝

PR #29, opened 2016-06-01
Feature/ux 295 - migrate Table to JDL

cliffmeyers:
Summary of changes
* This is mostly a straight port of the Table component in blueocean-dashboard. The only API change is that "fixed" table-layout is opt-in. I will update blueocean-plugin accordingly in a separate PR.
* Also added some nifty support to Storybook's webpack config so we can bundle CSS with our storybooks - which was useful to illustrate some of the class usage for this component.


src/js/components/Table.jsx
```diff
@@ -0,0 +1,60 @@
+import React, { Component, PropTypes } from 'react';
+
+const { array, node, string } = PropTypes;
+
+export class Table extends Component {
```
cliffmeyers: I will add some proper docs/comments to the class / component tomorrow.

src/js/stories/table.css
```diff
@@ -0,0 +1,11 @@
...
+    color: #ff0000;
+}
+
+.blue {
+    color: #0000ff;
+}
+
+.green {
+    color: #00ff00;
+}
```
sophistifunk: I really don't like having storybook-specific CSS, nor bringing in all the magic in the accompanying loaders. The point of storybooks should be to demonstrate how things will actually be, and we've got styles in the JDL already. There may be a good use case for this that I haven't thought of though, so let me know if that's the case.

src/js/components/index.js
```diff
@@ -10,8 +10,8 @@ export {WeatherIcon} from './weather-icon';
 export {Page} from './page';
 export {GlobalHeader, GlobalNav} from './global-header';
 export {PageHeader, Title, PageTabs, TabLink} from './page-header';
-export {Table} from './table';
 export {LogConsole} from './log/index';
+export {Table} from './Table';
```
sophistifunk: I was concerned about this case change, but I just tested merging this onto a throw-away clone master and it worked correctly on OSX so it's OK in this case, git was smart enough to do a delete then save, instead of a NOP rename for once.

michaelneale:
looks good in storybook, but having styles for storybook is perhaps not quite right as @sophistifunk points out. Ideally it would look like it does in blue ocean ?

cliffmeyers:
@michaelneale @sophistifunk the Table supports the ability to apply classNames on the TH elements, so I wanted to illustrate those classes/styles being applied in storybook. Generally I wouldn't see a need to add them here but I thought it might be useful here.

sophistifunk:
We don't actually have any alternative TH classes to test against though? I see the use-case, but maybe just add a `<style>` block to `.storybook/head.html`? I'm really reluctant to let the practice of `require('')`ing CSS as if were JavaScript get a toe in the JDL door, as I find it to be really abusing the `require('')` semantics. I suspect I'm going to get outvoted on this point though, as it seems to be one of the things all the cool kids are doing :"(

Anyway, I've said my piece; if you're sure this is the right way to go then merge away, and it's on me to think up something better. I'm not even your real dad ;-)

cliffmeyers:
@sophistifunk I went ahead and rolled back the webpack magic and added a few testing CSS classes to head.html as you suggest. If I run into this use case in the future, I will dig through the JDL theme and see what I can apply from there.

PR #31, opened 2016-06-01
feature/ux-311: use semantic "time" element for ReadableDate

cliffmeyers:
Summary of changes:
* Use time element and populate dateTime attribute
* Fix silly propTypes warning that kept spamming console in totally valid scenarios like a null/undefined value
* Added some storybooks


sophistifunk:
Looks good 🍻

i386:
@cliffmeyers when the recent batch of visual JDL changes in could you publish and update blue ocean please?

michaelneale:
🐝

i386:
@cliffmeyers to improve the readability of the precision format could we do something like:
* For dates falling in the same year `Apr 12 5:30pm UTC`
* For dates falling in a different year `Apr 12 2004 5:30pm UTC`
* Always convert the timezone to UTC (this might be nicer todo on the server – WDYT @vivek / @imeredith ?)

http://uxmovement.com/content/absolute-vs-relative-timestamps-when-to-use-which/

cliffmeyers:
@i386 to clarify, are you suggesting that we replace the relative date/time values with the absolute date time format? Or make the latter available in the tooltip?

Regarding dates in JSON, it looks like all of our dates are UTC except for the changeset objects, which use the timezone of the computer where the commit was made. If we added some server side logic to adjust that to UTC, then we'd be well-positioned to handle this cleanly in the UI.

vivek:
@i386 @cliffmeyers Server sends the time format in ISO-8601 format, that includes timezone information, for example, "2016-06-01T08:23:33.058-0700". If you want it in UTC format as: "2016-06-01T15:23:33.058 UTC", its small change on the server side and can be done easily. 

@cliffmeyers I just checked, change set timestamp is in iso-8601 as well, just like all other timestamp generated by server side. Can you send the JSON object for change set?

cliffmeyers:
@vivek the only property that I don't see in ISO-8601 is the "date" field from the changeset object (see below) which also expresses the local time zone of the commit. I didn't notice at first that we also have a "timestamp" field which is in UTC, which is great. 

```
"changeSet": [
      {
        "_class": "io.jenkins.blueocean.service.embedded.rest.ChangeSetResource",
        "author": {
          "_class": "io.jenkins.blueocean.service.embedded.rest.UserImpl",
          "email": "cmeyers@cloudbees.com",
          "fullName": "cmeyers",
          "id": "cmeyers"
        },
        "affectedPaths": [
          "src/js/components/TimeDuration.jsx"
        ],
        "commitId": "9f3ba4bf0f8d8392fa1aafc7e762119eb22dc4ff",
        "comment": "[UX-297] simplify parseInt logic\n",
        "date": "2016-06-01 09:52:26 -0400",
        "id": "9f3ba4bf0f8d8392fa1aafc7e762119eb22dc4ff",
        "msg": "[UX-297] simplify parseInt logic",
        "paths": [
          {
            "editType": "edit",
            "file": "src/js/components/TimeDuration.jsx"
          }
        ],
        "timestamp": "2016-06-01T13:52:26.000+0000"
      }
    ],
```

I am not sure if we should leave date as-is or if we even need it. I think for any date values we want to display in the UI as long as they are all in UTC that's good, so we can keep any timezone conversion logic in the UI simple and consistent.

vivek:
@cliffmeyers Oh, ok, so that 'date' field is coming from underlying model and should be ignored. ChangeSet comes from extension and returned as it is, this is side effect of using underlying extensions but no easy way out of it. So 'date' field should be treated as unknown field and 'timestamp' should be used. Jenkins model in general is not consistent about returning date string in standard format (ISO 8601).

cliffmeyers:
@vivek makes sense! thanks

michaelneale:
:bee:

PR #33, opened 2016-06-01
feature/UX-476: add tests and lint to JDL

cliffmeyers:
Summary of changes
* Added mocha, chai and enzyme with basic setup in gulp
* One sample test for ReadableDate using enzyme
* Added gulp-eslint and ruleset definition from blueocean and blueocean-dashboard


gulpfile.js
```diff
@@ -79,14 +85,30 @@ gulp.task("watch-styles", ["default"], () => {
 // Default to clean and build
 
 gulp.task("default", () =>
-    runSequence("clean", "build"));
+    runSequence("clean", "test", "build"));
```
cliffmeyers: I've excluded "lint" from the default build for now since there are 40 errors to clean up. Will do that in a second pass on this PR.

npm-shrinkwrap.json
```diff
@@ -1,6 +1,6 @@
 {
```
cliffmeyers: Our existing shrinkwrap config had enzyme locked to 2.2.0. The `containsMatchingElement` and related methods are only available in 2.3.0, so I ticked up that dep and rebuilt shrinkwrap. Looks like it hadn't been run in a very long time as there were lots of changes.

test/js/components/ReadableDate-spec.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { assert } from 'chai';
+import { shallow } from 'enzyme';
+
+import { ReadableDate } from '../../../src/js/components';
```
cliffmeyers: Any idea is there's a better way to handle this pathing? It feels kinda kludgy.

test/js/components/ReadableDate-spec.js
```diff
@@ -0,0 +1,44 @@
...
+    it("renders 'an hour ago' with current date minus 60m", () => {
+        const now = new Date();
+        const hourAgo = new Date(now.getTime() - 1000*60*60);
+
+        const wrapper = shallow(<ReadableDate date={hourAgo.toISOString()} />);
+
+        assert.equal(wrapper.containsMatchingElement(
+            <span>an hour ago</span>
+        ), true);
+    });
```
cliffmeyers: `containsMatchingElement` is really handy when you need to check simple things in the DOM.

cliffmeyers:
I wanted to get this PR up now for feedback on the mocha and eslint setups and enzyme usage. I still have a few more tasks I plan on to finish up tomorrow:

1. Fix all lint errors
2. Add "lint" into the default build so CI fails if lint doesn't pass
3. Write tests for several more components, just to raise our baseline a bit

cliffmeyers:
Also if anyone has thoughts on the best approach to make the test results useful to Jenkins, please let me know. I figured that this package might be worth looking at:
https://www.npmjs.com/package/mocha-junit-reporter

.eslintrc
```diff
@@ -0,0 +1,53 @@
...
+  "env": {
+    "browser": true,
+    "es6": true,
+    "node": true
+  },
+  "parser": "babel-eslint",
+  "rules": {
+    "eol-last": 1,
+    "no-unused-vars": [2, {"varsIgnorePattern": "^React$"}],
+    "max-len": [1, 120, 4],
```
michaelneale: 120 char lines? (much nicer than 80!)

michaelneale:
This looks like a nice cleanup to my eyes.

.eslintrc
cliffmeyers: These rules should be verbatim what we have in blueocean-plugin and dashboard right now. 120 width seems reasonable to me since it's not 1995 and we're not on 15" CRT's ;)

test/js/components/ReadableDate-spec.js
sophistifunk: At `${pastJob}` we just put the tests in the same directory as the regular code, and then skipped over it when disting / bundling. Personally I much prefer that method, so if you're happy to change the globs in the gulpfile to skip everything that ends in -spec, I'm happy to go that way. I don't much care for the `../../../../../../../../../../../foo/bar/baz/Widget.js` imports either :D

gulpfile.js
sophistifunk: I say fuck eslint, unless we remove formatting bullshit. "Lint" means "look for logic errors" not "bitch at the programmer until the code fits on a printer from 1978"

.eslintrc
sophistifunk: If it's not logic, it doesn't belong in a lint file. If you want to be a nazi and dictate the source format, use a beautifier. Also the existing xml formatting in our current setup in blueocean is *all fucked up*

tfennelly:
Looks good to me. Thanks Cliff 🐝

test/js/components/ReadableDate-spec.js
cliffmeyers: Yeah that sounds like a good approach to me. I know there's some more tweaks to come in testing, so I'll try to work that in soon.

.eslintrc
cliffmeyers: I tried to avoid any eslint rule debates by using exactly what was already defined in dashboard :P

cliffmeyers:
Going to merge this now. We were already living by the lint rules in dashboard, so I feel like we should at least be consistent across projects, even if it's not ideal. It's not enabled in the build either (at least, not yet) but I'd like to get the unit test stuff in so we can raise the bar for testing in the JDL.

PR #35, opened 2016-06-02
feature/ux 476 - delinting and some tests

cliffmeyers:
Summary of changes:
* Delinting: there are a handful of warnings left over but they don't fail the build
* Added unit tests for some of the simpler text components - hopefully we can add some for the others once we touch them again

@reviewbybees


michaelneale:
Nice housekeeping. Always sad when I see that instead of fixing "let" in js they introduced "const" (or maybe it is the reverse, in any case there are 300% more ways to assign a value than there used to be in js). 

🐝  !

PR #36, opened 2016-06-03
LogConsole bug fixes

cliffmeyers:
Summary of changes:
* Fix issue with selecting line numbers
* Fix issue with line numbers not displaying past ~40
* Make the LogConsole render in chunks, so it doesn't jank the animation of the modal sliding in

@reviewbybees


michaelneale:
this looks like a small enough change in short term - is that ok @sophistifunk @scherler ? 

🐝

sophistifunk:
LGTM

cliffmeyers:
I'll merge this in and then iterate with @i386 further if we need to. The downside to this impl is that while the logs are being loaded it janks the vertical scrollbar a bit. So if you have a massive log file and immediately try to scroll, it may be a finicky for a sec or two. I think this is still a good short-term fix given that it was super easy to implement and fixes the animation which looked quite awful in its that state.

PR #38, opened 2016-06-07
feature/ux 531- button styles

cliffmeyers:
Summary of changes:
* New styles
* Scrabbed default button style that was giving ugly gray buttons. I can't imagine we'd ever want them to look like that, so now default and primary are the same thing.
* Added storybook to illustrate styles

@reviewbybees


cliffmeyers:
Storybooks:

ux-531-buttons-default

ux-531-buttons-inverse

cliffmeyers:
@i386 added some screenies of the storybooks for your reference / approval.

i386:
LGTM!

michaelneale:
🐝

PR #40, opened 2016-06-08
experiment/use karma ci

cliffmeyers:
Summary of changes
* Fix an issue where karma had wrong extension in glob
* Polyfill Math.log10 (broke in Phantom, but also would have broken in IE)
* Add the mocha reporter
* Set up karma as default test runner from gulp


tfennelly:
LGTM 🐝

michaelneale:
🐝

PR #41, opened 2016-06-10
#UX-412 ResultItem for logs, tests

sophistifunk:
UX-412 Developer sees the log broken down by step -- ResultItem component

 * Add ResultItem component, styles, storybook
 * Some tangentially related fixes
 * Fix Flow errors
 * Switch gulp watch and watch-styles to not run tests or eslint

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node
+FROM node:5.11.1
 
 RUN useradd jenkins --shell /bin/bash --create-home
 USER jenkins
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,7 @@ node {
      stage "Prepare environment"
        deleteDir()
        checkout scm
+     stage "Build container"
        def environment  = docker.build 'cloudbees-node'
 
        environment.inside {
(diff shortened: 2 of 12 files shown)


sophistifunk:
Not sure what's wrong with FSEvents, but I've been getting a lot of angry reports from V8 about stuff soon to be removed that's probably related :-/

michaelneale:
@sophistifunk not sure why but a failure running the build: 

sh: 1: node-pre-gyp: not found
npm info lifecycle fsevents@1.0.11~install: Failed to exec install script

could be environmental

michaelneale:
@sophistifunk if you think it is a node version thing: https://hub.docker.com/_/node/ has the versions available, Putting those version tags at the top of the Dockerfile and updating this PR will cause the build to run with that version of node available.

i386:
So glad to see this landing 🎉👌

sophistifunk:
@i386 you and me both 😆

michaelneale:
@i386 sadly when I see spurious failures of npm I see days ahead of yak shaving :(

michaelneale:
OK looking good. 

@sophistifunk sorry I had to muck with this, let me explain: 

1) can't invoke "npm run gulp" right now - as paths have "%" in them (due to naming of pull request). known problem with npm (Jenkins will soon fix to not use things like that in paths)
2) it seems on linux this build is not stable with node 6, so I set the version in the Dockerfile to 5.11.1, which seems to make it reliable.

PR #42, opened 2016-06-10
[UX-512] updates to Pipeline Result to allow for linked elements

cliffmeyers:
Related to issue # UX-512.

Summary of changes
* Apply white text styles for links in subheader and pipeline result header
* Add ability to specify callback funcs as props on pipeline result


i386:
LGTM

PR #44, opened 2016-06-13
Misc design tweaks that get style closer to mockups

i386:
@reviewbybees

### DIFF
--- a/less/theme.less
+++ b/less/theme.less
@@ -26,10 +26,12 @@ html {
 
 body {
     color: @text-color;
-    text-rendering: optimizeSpeed;
-    font-feature-settings: "kern" on, "liga" off;
-    -moz-font-feature-settings: "kern" on, "liga" off;
-    //TODO: Do we really want this for everything? -webkit-font-smoothing: antialiased;
+  text-rendering: optimizeSpeed;
+  font-feature-settings: "kern" on, "liga" off;
+  -moz-font-feature-settings: "kern" on, "liga" off;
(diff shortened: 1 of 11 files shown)


less/theme.less
```diff
@@ -26,10 +26,12 @@ html {
...
     color: @text-color;
-    text-rendering: optimizeSpeed;
-    font-feature-settings: "kern" on, "liga" off;
-    -moz-font-feature-settings: "kern" on, "liga" off;
-    //TODO: Do we really want this for everything? -webkit-font-smoothing: antialiased;
+  text-rendering: optimizeSpeed;
+  font-feature-settings: "kern" on, "liga" off;
+  -moz-font-feature-settings: "kern" on, "liga" off;
+  // Want to change this? Pick a fight with i386
+  -webkit-font-smoothing: antialiased;
```
sophistifunk: worse-than-hitler

less/theme.less
```diff
@@ -642,6 +664,10 @@ code.hash {
     a {
         color: white;
     }
+
+    h1 {
+      font-size: 1.8rem;
```
sophistifunk: Not `@font-size-h1`?

sophistifunk:
LGTM 🐝

less/theme.less
i386: haha

less/theme.less
i386: Nah mate – the standard H1 is a bit too big to use in the nav.

PR #48, opened 2016-06-14
bug/status-indicators-lost-fill

cliffmeyers:
Summary of changes:
* CSS class name was updated but component needed its class names updated too


cliffmeyers:
Simple change, self-merging.

PR #50, opened 2016-06-15
[JENKINS-35690] tweak to LiveStatusIndicator

cliffmeyers:
* Rework LiveStatusIndicator to use ISO-8601 string for "startTime" rather than millis, to ensure stabler date arithmetic

@reviewbybees


i386:
LGTM

PR #51, opened 2016-06-15
Fix a couple of issues in PipelineGraph with duplicate node names

sophistifunk:
* Allow duplicate stage/node names in PipelineGraph
* Return id as well as name for onNodeClick
* Update stories

@reviewbybees


scherler:
as soon build went through :bee:

PR #52, opened 2016-06-15
Add expand-in transition to ModalView

kzantow:
A new style for the modal transition that makes me much happier.

@reviewbybees esp. @i386


i386:
LGTM

PR #53, opened 2016-06-15
Minor tweaks to ResultItem component

kzantow:
... to be styled a bit better and not hardcode the right-aligned text type as a duration.

image


kzantow:
@reviewbybees esp. @sophistifunk @i386

i386:
LGTM 💯

PR #56, opened 2016-06-16
feature/jenkins 35912 - improve status indicator to support "modal header" use case

cliffmeyers:
Related to issue # JENKINS-35912

Summary of changes:
* Enhance Live/StatusIndicator with a "noBackground" property that hides the circle and scales up glyph so it fills the box (to ease layout)
* Update the Pipeline Result to use the new indicator in this state
* Some margin and color tweaks
* Temporarily stripping out some Flow checking to fix numerous errors

### DIFF
--- a/src/js/components/status/LiveStatusIndicator.jsx
+++ b/src/js/components/status/LiveStatusIndicator.jsx
@@ -119,7 +119,7 @@ LiveStatusIndicator.propTypes = {
     percentage: number,
     width: string,
     height: string,
-    chromeless: bool,
+    noBackground: bool,
     startTime: string,
     estimatedDuration: number,
 };
--- a/src/js/components/status/LiveStatusIndicator.jsx
+++ b/src/js/components/status/LiveStatusIndicator.jsx
@@ -2,8 +2,6 @@ import React, {Component, PropTypes} from 'react';
 import moment from 'moment';
 import { StatusIndicator, decodeResultValue } from './StatusIndicator';
 
-const {bool, number, string} = PropTypes;
-
 /**
  * LiveStatusIndicator is a wrapper around StatusIndicator that allows
  * for an in-progress status to self update.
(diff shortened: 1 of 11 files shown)


cliffmeyers:
This one might be good to have @sophistifunk take a look.

src/js/components/status/LiveStatusIndicator.jsx
```diff
@@ -1,25 +1,8 @@
...
-    queued: 'queued',
-    unstable: 'unstable',
-    aborted: 'aborted',
-    not_built: 'not_built',
-    unknown: 'unknown'
-};
-
-// Enum type from const validResultValues
-export type Result = $Keys<typeof validResultValues>;
+const {bool, number, string} = PropTypes;
```
sophistifunk: Please don't do this in JDL code, it shadows Flow type names

src/js/components/status/StatusIndicator.jsx
```diff
@@ -4,7 +4,7 @@ import React, {Component, PropTypes} from 'react';
 import SvgSpinner from './SvgSpinner';
 import SvgStatus from './SvgStatus';
 
-const {number, string} = PropTypes;
+const {bool, number, string} = PropTypes;
```
sophistifunk: As above

src/js/components/status/LiveStatusIndicator.jsx
cliffmeyers: Just to clarify, we want to use PropTypes.bool rather than importing it as a const?

sophistifunk: Yup. Otherwise bad things will almost certainly happen, as `React.PropTypes.*` are run-time values (and functions), which can be used as types in Flow. Either when they're constructor functions, or for enum unions; a Flow union could be `string | number` being flow types, or it might be `-1 | 0 | 1` being JS values

cliffmeyers: Ok thanks for clarifying. These are fixed in 396fbff

sophistifunk:
bbm-bueno

PR #61, opened 2016-06-18
Jenkins-35820 PipelineGraph selection

sophistifunk:
Add the ability to select a node in the pipeline graph via UI or props, as well as some initial visual selection feedback

@reviewbybees


sophistifunk:
screen shot 2016-06-18 at 11 28 58 am

i386:
That works :) 👍

PR #65, opened 2016-06-22
bug/JENKINS-36056 - fix bug with incorrect status indicator

cliffmeyers:
Related to issue # JENKINS-35056

Summary of changes:
* Handle case where estimatedDuration is negative

@reviewbybees


cliffmeyers:
Simple change, so merging w/o review.

PR #66, opened 2016-06-23
bug/jenkins 36128 - durations not updating

cliffmeyers:
Related to issue # JENKINS-36128

Summary of changes:
* Enhance TimeDuration component to support live updating. Using a much less aggressive 30s update interval. Long term I'd like to improve this further to use an intelligent setInterval period (similar to exponential backoff).
* Updated PipelineResult to support live updating. ReadableDate always live updates; TimeDuration only updates while the result is actually running.

@reviewbybees


michaelneale:
:bee:

PR #67, opened 2016-06-24
bug/jenkins 36102 - tabs clipped safari

cliffmeyers:
Related to issue # JENKINS-36102

Summary of changes
* Fix a recent regression where the Run Details tabs were being clipped in Chrome / Firefox
* Fix a longer-standing bug where the tabs were severely clipped in Safari

@reviewbybees

### DIFF
--- a/less/theme.less
+++ b/less/theme.less
@@ -832,6 +832,12 @@ footer {
     }
 }
 
+@supports (overflow:-webkit-marquee) and (justify-content:inherit) {
+    .dialog .header .page-tabs {
+        margin-bottom: 22px;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


less/theme.less
```diff
@@ -836,7 +829,13 @@ footer {
 
     a {
         letter-spacing: 1.1px;
-        margin-bottom: -4px;
```
cliffmeyers: `.page-tabs` styles were tweaked in an earlier tab to not use negative margins, so these rules above are not necessary, and in fact were causing some clipping issues on Run Details.

i386:
This LGTM

PR #68, opened 2016-06-30
feature/JENKINS-35828 - status colors

cliffmeyers:
Related to issue #JENKINS-35828

Summary of Changes:
* Add some generic classes for setting fill, background-color or color of an element in either normal or light theme color, based off result/status name. This should help us avoid having to remap our brand colors to numerous different classes.


sophistifunk:
ohmyyes

PR #70, opened 2016-07-01
Feature/jenkins 35828 status colors

cliffmeyers:
Related to issue # JENKINS-35828

Summary of changes:

* Remove baked-in margin from Favorite that will cause positioning issues when it needs to be flush-right
* Fix a bug where the "noBackground" variant of LiveStatusIndicator was not correctly rendering in the queued and not_built states
@reviewbybees


tfennelly:
🐝

PR #73, opened 2016-07-06
feature/JENKINS-35828 - favorite tweaks

cliffmeyers:
Related to issue #JENKINS-35828 

Summary of changes:
* Add new theme option for Favorite (white fill when selected)
* Added storybook

Note: Added raw SVG file just for reference

@reviewbybees


src/js/components/favorite/Favorite.jsx
```diff
@@ -15,8 +15,8 @@ export class Favorite extends Component {
     };
 
     static defaultProps = {
-        darkTheme: false,
-        checked: false
+        checked: false,
+        className: ''
```
kzantow: ?

less/components/favorite.less
```diff
@@ -32,7 +32,11 @@
...
+  }
```
tfennelly: Where are these data images coming from? I think LESS has some functions for generating this into the CSS from an image file. Might be easier to maintain.

src/js/components/favorite/Favorite.jsx
cliffmeyers: There are three themes now, so instead of the boolean I just have the style applied via a class. If you specify the `className` property on a React component it's actually ignored by default, so you need to re-expose an explicit property.

src/js/components/favorite/Favorite.jsx
```diff
@@ -58,7 +52,7 @@ export class Favorite extends Component {
 }
 
 Favorite.propTypes = {
-    darkTheme: PropTypes.bool,
     checked: PropTypes.bool,
+    className: PropTypes.oneOf(['', 'dark-yellow', 'dark-white']),
```
kzantow: I'm not at all fond of building in themes to components with colors as their names. It's unclear to me why any of these are using themes anyway and we're not just switching some higher-level css file?

src/js/components/favorite/Favorite.jsx
tfennelly: Blank `className` is intentional?

kzantow:
:bee: but I think this theming should be done a completely different way (I know this isn't the first example)

tfennelly:
🐝

less/components/favorite.less
cliffmeyers: I am generating the Base64 more or less by hand from the SVG's from James. And you're right, Less has `data-uri` which will do the embed for us. Ideally we'd just let the browser load the SVG asset directly but I seem to recall running into a problem with that while developing the component.

src/js/components/favorite/Favorite.jsx
cliffmeyers: I'm not really thrilled with it either. Are you proposing that the style should be applied via a class on a parent container?

less/components/favorite.less
cliffmeyers: I actually externalized these so the browser will load the SVG's directly. In commit d4702d3

PR #74, opened 2016-07-08
feature/JENKINS-35831 - favorite fix

cliffmeyers:
Related to issue # JENKINS-35381.

Summary of changes:
* Fix a bug in Favorite component that caused it to not check correctly if "checked" property changed after initial render


cliffmeyers:
Need to sort out a few other issues before i can verify this is the correct fix. Closing for now.

PR #75, opened 2016-07-11
feature/JENKINS-35831 - favorites fix

cliffmeyers:
Related to issue JENKINS-35831

Summary of changes:
* Fix a bug in Favorite component that caused it to not check correctly if "checked" property changed after initial render


cliffmeyers:
This is a simple fix; self-merging

PR #76, opened 2016-07-12
[feature/JENKINS-35886] Remove check whether expanded was changed sinc…

scherler:
…e it collapses in case of links

see https://issues.jenkins-ci.org/browse/JENKINS-35886.

### DIFF
--- a/src/js/components/ResultItem.jsx
+++ b/src/js/components/ResultItem.jsx
@@ -51,11 +51,9 @@ export class ResultItem extends Component {
             const statusGlyph = getGlyphFor(resultClean);
             this.setState({resultClean, statusGlyph});
         }
-        if (props.expanded != this.props.expanded) {
-            // check whether we want to change the state or whether we already are in the correct state
-            if (props.expanded !== this.state.expanded) {
-                this.toggleExpanded();
-            }
+        // check whether we want to change the state or whether we already are in the correct state
+        if (props.expanded !== this.state.expanded) {
+            this.toggleExpanded();
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/js/components/ResultItem.jsx
```diff
@@ -51,9 +51,6 @@ export class ResultItem extends Component {
             const statusGlyph = getGlyphFor(resultClean);
             this.setState({resultClean, statusGlyph});
         }
-        if (props.expanded!=this.props.expanded) {
```
kzantow: removing this will cause the callbacks not to fire: `onExpand` & `onCollapse`, that seems like a :bug:

scherler: On props change yes, that is the intend of the change. 

The component use the state to keep track on that events see toggleExpanded

```
const expanded = !this.state.expanded;
```

The problem is that above code is, it triggers  this.toggleExpanded(); and not check the new value of the expanded from nextProps but simply toggles the current expanded in the state. Now in our use case we have expanded via onExpand and clicking on a line in the log will add the hash to the current location. That will change the prop of the step in which the log resist and will collapse it which is not what it should do.

...but writing this explanation I think I have an alternative approach. Cheers @kzantow  for being 
rubber-duck_0

kzantow: it was checking the incoming `props.expanded != this.props.expanded`. Why not just change it to `this.state.expanded != props.expanded`? And thank you for being:
image

kzantow:
:bee: AFAICT, the logic and/or failure cases are still a bit fuzzy to me. probably should have at least a negative/positive test for this...

michaelneale:
🐝

PR #77, opened 2016-07-12
[JENKINS-36299] add empty state view handling for Icon types

kzantow:


src/js/components/EmptyStateView.jsx
```diff
@@ -37,6 +51,6 @@ export class EmptyStateView extends Component {
 
 EmptyStateView.propTypes = {
     children: PropTypes.node,
-    iconName: PropTypes.oneOf(['branch','goat','shoes']),
+    iconName: PropTypes.string,
```
michaelneale: :/

kzantow: because... ? this allows any material icon type, I don't think it's better to hardcode all this stuff further

kzantow: .. to me it's insanity that this is hardcoded at all, in multiple places

michaelneale: fair enough. My only remaining objection is the word "goat" is no longer found in the codebase.

michaelneale:
any op to add some tests quality to this? otherwise 🐝

src/js/components/EmptyStateView.jsx
kzantow: 'goat' is still found in other places (the multiple hardcoding I mentioned). :)

kzantow:
@michaelneale I did add a storybook entry for this, tests are pretty skimpy for this module

michaelneale:
🐝

src/js/stories/emptystateview.jsx
```diff
@@ -33,6 +34,14 @@ function contentWithNarrowImage() {
     );
 }
 
+function contentWithIcon() {
```
scherler: I am not really a fan of multiple component in one file and consider the pattern you show as problematic. Why "hide" a React component in a function, I now it is valid but not really re-usable

scherler:
other then my comment :bee:

src/js/stories/emptystateview.jsx
kzantow: I'm confused. This is the storybook, I'm just following the existing pattern.

michaelneale: yeah I think its fine for storybook

michaelneale:
:bee:

PR #78, opened 2016-07-14
adds junit test output to mocha

i386:
**Decription**
This change adds junit test output to mocha so we can use our super new test reporting feature!

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explaination given

@reviewbybees


michaelneale:
LGTM if no objections

i386:
@tfennelly could we do the same for all the blueocean modules?

cliffmeyers:
Awesome! 🐝

PR #79, opened 2016-07-15
[JENKINS-36701] add some validation to end of the gulp build

cliffmeyers:
**Decription**
* JDL was published without octicons because npm install hadn't been run recently on publisher's machine. To defend against that, I added some validation to end of the gulp build, asserting that a few important directories are there
* Tested locally, no easy way to write a test for the build :P

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code
- [x] Verified that the appropriate tests have been written or valid explaination given

@reviewbybees

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,7 @@
  */
 
 const gulp = require('gulp');
+const gutil = require('gulp-util');
 const fs = require('fs');
 const sourcemaps = require('gulp-sourcemaps');
 const babel = require('gulp-babel');
@@ -209,10 +210,11 @@ gulp.task("validate", () => {
     ];
 
     for (const path of paths) {
-        fs.stat(path, (err) => {
-            if (err != null) {
-                throw err;
-            }
-        });
+        try {
+            fs.statSync(path);
+        } catch (err) {
+            gutil.log('Error occurred during validation; see stack trace for details');
+            throw err;
+        }
     }
 });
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -197,3 +198,21 @@ gulp.task("copy-licenses-octicons", () =>
...
+
+gulp.task("validate", () => {
+    const paths = [
+        config.less.dest,
+        config.copy.fonts.dest,
+        config.copy.octicons.dest,
+    ];
+
+    for (const path of paths) {
+        fs.stat(path, (err) => {
```
sophistifunk: We might want to use statSync() here instead? Otherwise I don't think gulp will notice the callback throwing, even if it happens before gulp finishes.

michaelneale:
:bee: although I think @sophistifunk had a question about sync and not actually noticing the error.

sophistifunk:
It smells like the kind of thing we may not be able to *prove* is a problem, and it'll just come up on a slow machine, depending on the phase of the moon / cosmic rays. I'd like to see it changed to sync just-in-case, as a blocking FSTAT isn't really going to hurt anything

cliffmeyers:
@sophistifunk agree, your suggestion is very sensible, and I will definitely implement it. I just hadn't responded because some other tasks were my focus last week. Should get to this in the next day or so.

cliffmeyers:
@sophistifunk fixed in commit 4451a8s

PR #81, opened 2016-07-25
[JENKINS-36169] Fix offset for logs until <10k. For bigger logs you s…

scherler:
**Decription**
General bug ATM with line padding >999 (log line moves to the right)

BEFORE
screenshot from 2016-07-25 14-25-42

AFTER
screenshot from 2016-07-25 14-24-25

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explaination given

@reviewbybees 

For bigger logs >10k you still see the offset bug. @i386 what should we do in this cases?

screenshot from 2016-07-25 14-24-39

### DIFF
--- a/less/components/result-status.less
+++ b/less/components/result-status.less
@@ -220,7 +220,6 @@
     background-color: @pre-bg;
     color: @pre-color;
     padding: 0.5em;
-    border-left: solid 28px darken(@pre-bg, 5%);
     margin-bottom: 1px;
 }
 
--- a/less/components/result-status.less
+++ b/less/components/result-status.less
@@ -221,6 +221,7 @@
     color: @pre-color;
     padding: 0.5em;
     margin-bottom: 1px;
+    border-left: solid 28px darken(@pre-bg, 5%);
 }
 
 // Expand animation
--- a/less/theme.less
+++ b/less/theme.less
@@ -592,7 +592,7 @@ code p::before {
     color: @gray-light;
     content: counter(line);
     float: left;
-    min-width: 20px;
+    min-width: 30px; // scales < 10k then you will see an offset
     white-space: nowrap;
     padding-right: 1em;
     padding-left: 10px;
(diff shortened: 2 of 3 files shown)


i386:
@scherler could we make the gutter grow according to the number of characters of the line number and align the numbers to the right?

scherler:
@i386 you mean changing them on the go?

How would that perform in karaoke mode, we do not know the lines we are about to render. Wouldn't that look weird that "suddenly" (every 100/1000) the log body got moved a bit to the right?

i386:
@scherler how about leaving space for 5 digits and let it auto expand from that point on? That way it only sucks for an instant when it goes over 100k lines

scherler:
done, can I get a :bee:, please?
screenshot from 2016-07-27 13-28-28

sophistifunk:
🐝

sophistifunk:
This will be our log console by the time we get a chance to fix it 
server-room

less/components/result-status.less
```diff
@@ -220,7 +220,6 @@
     background-color: @pre-bg;
     color: @pre-color;
     padding: 0.5em;
-    border-left: solid 28px darken(@pre-bg, 5%);
```
kzantow: does this break the gutter on the test results?

less/theme.less
```diff
@@ -589,10 +589,10 @@ code a:hover {
 }
 
 code p::before {
+    background-color: @brand-grey;
     color: @gray-light;
     content: counter(line);
-    float: left;
-    min-width: 20px;
+    min-width: 35px; // scales < 100k then you will see an offset
```
kzantow: maybe use `em` or `rem` ?

less/components/result-status.less
scherler: I am not sure, @i386 asked that this should be removed. I had an override before for the logConsole. If you you need this I can switch back to the override

PR #84, opened 2016-07-29
[JENKINS-36582] fix a bug in LiveStatusIndicator where the estimated …

cliffmeyers:
**Description**

**Submitter checklist**

**Reviewer checklist**
- [x] Run the changes and verified the change matches the issue description
- [x] Reviewed the code
- [x] Verified that the appropriate tests have been written or valid explaination given

@reviewbybees 

…progress would not draw correctly if it received new props that triggered a running state

### DIFF
--- a/src/js/components/status/LiveStatusIndicator.jsx
+++ b/src/js/components/status/LiveStatusIndicator.jsx
@@ -83,14 +83,15 @@ export class LiveStatusIndicator extends Component {
                 this._updateProgress();
             }, 1000);
 
-            this._updateProgress();
+            this._updateProgress(props);
         }
     }
 
-    _updateProgress() {
+    _updateProgress(props) {
         const now = moment();
         const elapsed = now.diff(this.startTime);
-        this.percentage = Math.floor(elapsed / this.props.estimatedDuration * 100);
+        const estimatedDuration = props && props.estimatedDuration || this.props.estimatedDuration;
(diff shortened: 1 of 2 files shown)


src/js/components/status/LiveStatusIndicator.jsx
```diff
@@ -80,17 +80,18 @@ export class LiveStatusIndicator extends Component {
...
+            this._updateProgress(props);
         }
     }
 
-    _updateProgress() {
+    _updateProgress(props: Props) {
         const now = moment();
         const elapsed = now.diff(this.startTime);
-        this.percentage = Math.floor(elapsed / this.props.estimatedDuration * 100);
+        const estimatedDuration = props && props.estimatedDuration || this.props.estimatedDuration;
```
sophistifunk: Looks like this will asplode if you've got a 0 for `estimatedDuration`? Maybe add a `|| elapsed` or something at the end?

sophistifunk:
Just one small question, but LGTM 🐝

src/js/components/status/LiveStatusIndicator.jsx
cliffmeyers: Reworked this a bit in 8f4554a to only calculate the percentage if we receive valid numbers, otherwise just put the spinner to indeterminate.

michaelneale:
:bee:

PR #86, opened 2016-08-18
Feature/jenkins 36199 toaster

cliffmeyers:
**Description**
- `Toaster` now receives an array of data objects as props and handles rendering of multiple toasts in a sensible way.
- Next step in separate PR is to add a `ToastService` to `blueocean-core-js` that clients can use to push in new Toasts, and to add `Toaster` to `blueocean-web` so its part of main Blue Ocean UI.
- Unfortunately the use of ReactCssTransitionGroup makes this hard to test w/ shallow renders; hopefully the storybook "counts" (but I know it really doesn't). Feel free to flog me for this and perhaps I can do some deep renders.

**Submitter checklist**
- [x] Instructions for reviewer:
 - You can load up the `Toaster` example in Storybook to see it in action.

**Reviewer checklist**
- [x] Reviewed the code
- [x] Verified that the appropriate tests have been written or valid explanation given

@reviewbybees

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 160
+
+[*.md]
+trim_trailing_whitespace = false
--- a/less/components/toast.less
+++ b/less/components/toast.less
@@ -1,59 +1,83 @@
+.toaster {
+    width: 100%;
+    height: 100%;
+
+    .toast {
+        margin-bottom: 10px;
+    }
+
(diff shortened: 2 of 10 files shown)


tfennelly:
🐝 LGTM @cliffmeyers 

Is there no way of adding a few simple unit tests for this? Why does `ReactCssTransitionGroup` complicate it?

cliffmeyers:
@tfennelly I added some tests for Toaster. If you use Enzyme's `shallow` then it treats the render like a true unit test, meaning it will not render any components nested inside of your component. Since ReactCssTransitionGroup is usually a top-level container in a component, it makes `shallow` pretty useless. `mount` can be finicky since it relies on jsdom but I was able to use the `render` method to some basic DOM assertions in a simple way. Going to go ahead and merge this now.

PR #87, opened 2016-08-25
[FIX JENKINS-36660] hooks for AT

scherler:
**Description**

This small change helps to match the nav tabs in AT, otherwise we need to match against URL which is more complicated in folder/multibranch projects

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@jenkins-cd/design-language",
   "jdlName": "jenkins-design-language",
-  "version": "0.0.71-unpublished",
+  "version": "0.0.71-thor2",
   "description": "Styles, assets, and React classes for Jenkins Design Language",
   "main": "dist/js/components/index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


michaelneale:
seems good to me - @tfennelly ?

scherler:
BLOCKS 

- jenkinsci/blueocean-acceptance-test#26

- https://github.com/jenkinsci/blueocean-plugin/pull/433

tfennelly:
🐝 insofar as I get them.

So ... would adding a class on an element in the page not help here and be easier to use ... seems like using URLs is going to be brittle, no?

scherler:
That is the change, we are using now a class and not the URL anymore, since building url in a sane way for multibranch in folder in AT is no fun only to match some tab

tfennelly:
@scherler duh .... now I see it ... sorry. Thanks.

tfennelly:
🐝  again , just in case there was a doubt.

kzantow:
:bee:

PR #94, opened 2016-09-07
Experiment/publish workflow

cliffmeyers:
**Description**
- Experimenting with formalizing the versioning and publishing process.
   - use 'npm run release' to do a real publish / release
   - use 'npm run release-beta' for a beta publish
- Sorry the diff sucks; needed to format package.json so it's consistent with all of our other formatting
- cc @tfennelly @scherler @kzantow @imeredith @sophistifunk @michaelneale 

**Submitter checklist**
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [ ] Appropriate unit or acceptance tests or explanation to why this change has no tests

**Reviewer checklist**
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

@reviewbybees

PR #96, opened 2016-09-16
Task/jenkins 37715 publish workflow

cliffmeyers:
**Description**
- This is more or less a replacement of #94 just with a slightly cleaner commit history. I've tested "npm run release" and "npm run release-beta" and they both work as expected (except for me site-publish fails because of some pub key config w/ github, but not really an issue with these changes).
- Added a script that performs a number of checks:
  - No ambiguous versions in deps or devDeps
  - No duplicates in devDeps that are already in deps
  - No deps in package.json that aren't in npm-shrinkwrap.json
- Also discovered that we probably need to stick with 2-space indent in package.json because any npm operation against that file (install --save/Dev, version, etc) will enforce that spacing and there's no sense fighting it. My bad for not realizing that sooner.
- See [JENKINS-37715](https://issues.jenkins-ci.org/browse/JENKINS-37715).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
cc @tfennelly @sophistifunk

tfennelly:
🐝 LGTM

tfennelly:
The likes of https://www.npmjs.com/package/release-it also look intersting

cliffmeyers:
I'm going to put this PR on-hold. I think @tfennelly 's suggestion of release-it may be a better long-term solution for us.

PR #97, opened 2016-09-26
(Hopefully) fix the issues caused by old node-sass transitive dependency

sophistifunk:
Don't ask me what particular change fixes it, but this *looks* like it's fixed :-/

@reviewbybees


michaelneale:
:bee: seems to build locally for me - but not sure how else to review it.

i386:
magic

PR #106, opened 2016-10-18
JENKINS-36288 JDL component for Popover

sophistifunk:
**Description**
- See [JENKINS-36288](https://issues.jenkins-ci.org/browse/JENKINS-36288).

**Submitter checklist**

**Reviewer checklist**
- [x] Run the changes and verified the change matches the issue description
- [x] Reviewed the code

@reviewbybees

Squashed commit of the following:
    feature/JENKINS-36288-popover * lint
    feature/JENKINS-36288-popover * MOAR FLOW
    feature/JENKINS-36288-popover * Flow
    feature/JENKINS-36288-popover * Add the little 'arrow' pointer, and fix the rest of the styles to match mockups
    feature/JENKINS-36288-popover * Dump storybook knobs, split it up into 4 stories instead
    feature/JENKINS-36288-popover * Initial import of class from testbed project, basic story

### DIFF
--- a/src/js/components/Popover.jsx
+++ b/src/js/components/Popover.jsx
@@ -1,0 +1,536 @@
+// @flow
+
+import React, { Component, PropTypes } from 'react';
+
+//--------------------------------------------------------------------------
+//
+//  Constants / types
+//
+//--------------------------------------------------------------------------
+
+const pollingInterval = 50; // ms
+const fixedAnimationDuration = 500; // ms
+
+const positionAbove = 'above';
+const positionBelow = 'below';
+const positionLeft = 'left';
+const positionRight = 'right';
+
+const positionValues = {
+    above: 'above',
+    below: 'below',
+    left: 'left',
+    right: 'right'
+};
+
+export type Position = $Keys<typeof positionValues>;
+
+export const positions: Array<Position> = Object.keys(positionValues);
+
+const stateInit = 'init'; // Created, not yet correctly positioned
+const stateStable = 'stable'; // Correctly positioned, all good
+const stateMoving = 'moving'; // Moving to new correct position
+
+const lifecycleStates = {
+    init: 'init',
+    stable: 'stable',
+    moving: 'moving'
+};
+
+export type LifecycleState = $Keys<typeof lifecycleStates>;
+
+type Props = {
+    position?: Position
+};
+
+//--------------------------------------------------------------------------
+//
+//  Helpers
+//
+//--------------------------------------------------------------------------
+
+//--------------------------------------
+//  Position enum
+//--------------------------------------
+
+function sanitizePosition(input) {
+
+    if (positions.indexOf(input) === -1) {
+        return positionAbove;
+    }
+
+    return input;
+}
+
+//--------------------------------------
+//  Animation easing
+//--------------------------------------
+
+// t: current time, b: beginning value, c: change In value, d: duration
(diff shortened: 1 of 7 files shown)


sophistifunk:
screen shot 2016-10-18 at 3 57 19 pm

michaelneale:
I cleaned node_modules and ran npm install.. but when I run the storybook I see: 

```
PropTypes is not defined
ReferenceError: PropTypes is not defined
    at Object.eval (webpack:///:105:15)
    at eval (webpack:///src/js/stories/PopoverStories.jsx:4165:2)
    at Object.<anonymous> (webpack:///src/js/stories/PopoverStories.jsx?d152:4165:2)
    at __webpack_require__ (http://localhost:9001/static/preview.bundle.js:556:30)
    at fn (http://localhost:9001/static/preview.bundle.js:87:20)
    at eval (http://localhost:9001/static/preview.bundle.js:4201:2)
    at Object.<anonymous> (http://localhost:9001/static/preview.bundle.js:4201:2)
    at __webpack_require__ (http://localhost:9001/static/preview.bundle.js:556:30)
    at fn (http://localhost:9001/static/preview.bundle.js:87:20)
    at loadStories (http://localhost:9001/static/preview.bundle.js:4057:2)
```

sophistifunk:
Should be fixed now

michaelneale:
Tried it out - LGTM. 

Code seems neat and easy to read 👍🏻
No tests (not sure how would test in unit testing - probably a few lines of code that could verify calculations, but wouldn't be the bulk...)

A question: the positioning is explicit - if you don't provide a position than `sanitizePosition` will default to `positionAbove` right? (heh - American spelling, I guess we have to). 

Are there any plans to have it automatically work that out based on some heuristic on positioning? (that calculation likely could be unit tested). I can see it does an adjustment to try and get it in the viewport: `Do a basic adjustment to make sure it's within the viewport if possible` - however, wouldn't it be better if it calculated it needed to be moved that much, it should change the `position` to something sane? Otherwise what we have now is that a bottom positioned popover, if the object is in the bottom left, will be shifted up with the little triangle pointing at the wrong thing. 

Looks like a good first pass, perhaps @cliffmeyers may have some feedback given he may be using it, at least on the api. 

Design wise looks great. Main thing is that automagic position problem (that could come later perhaps, I expect in the editor this will be needed)

src/js/components/Popover.jsx
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+    /**
+     Check for invalidated positioning, and re-calculate / move if necessary
+
+     Should only be called via window.requestAnimationFrame().
+     */
+    validatePositioning() {
+        if (!this.positioningValid) {
+            this.calculateAndSetPopupPosition();
```
michaelneale: Should this instead change the position to what it thinks is more appropriate (vs just translating)?

sophistifunk: Not sure what you mean by this?

sophistifunk:
There's room to change the way the "adjustment" works, and to move the little arrow indicator around, but it shouldn't change the preferred location, because it should go back to that if it's possible. 

IMO we should wait until James and Brody get some time to get a feel for it before we figure out the rules for more complicated re-positioning. Like most auto-layout problems, it's difficult to get right without a general constraint solver.

michaelneale:
Sounds good to me. :bee:

PR #110, opened 2016-11-04
Feature/pre dialog

sophistifunk:
**Description**

Bringing some updates for latest Flowtype and extracted ModalContainer component in from my dialog dev branch early in the hope of smaller more understandable commits and fewer merge issues next week.

- See [JENKINS-39371](https://issues.jenkins-ci.org/browse/JENKINS-39371).

**Submitter checklist**
- [X] Link to JIRA ticket in description, if appropriate.

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


sophistifunk:
Ping @cliffmeyers

src/js/components/Popover.jsx
```diff
@@ -2,22 +2,23 @@
...
+import {ModalContainer} from '.';
 
 export class Popover extends Component {
 
     //--------------------------------------
     //  User Interaction
     //--------------------------------------
 
-    modalScreenClicked:HTMLEventHandler = () => {
+    modalScreenClicked = () => {
```
cliffmeyers: Ah, I take this is another way of binding a class method's scope so that `this` is the class instance? Probably more performant (and readable) than using fat arrow within the `render` method?

sophistifunk: I think it's a wash in readability, but it's mostly so that the value is the same, otherwise every time render() is called you'll get a different value (!==) for the closure even if there's no locals bound, and so it's another thing react has to change in the DOM. I'd prefer to use a `@bind` annotation on a normal method, but that requires adding a bit more to our babel setup.

cliffmeyers: Cool, thanks for clarifying.

PR #111, opened 2016-11-07
Task/jenkins 39478 popover positioning

cliffmeyers:
**Description**
- Refactor positioning to be pluggable but preserve the simple API for the original `Popover` component.
- @sophistifunk curious about your input as to the API. Strategy pattern here may be overkill, although maybe it paves the way for other hooks we'd want to add later in the future such as dismissing, anim, etc.

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


sophistifunk:
Code looks fine, but I would definitely go with a function rather than a "strategy class", it just gives you a lot more flexibility, and IMO we need to try and avoid writing Java in Javascript syntax ;-)

cliffmeyers:
@sophistifunk cheers, I'll see if I can get something a bit simpler w/ just a pure func then.

sophistifunk:
Looks good 🐝

PR #113, opened 2016-11-09
JENKINS-39371 Dialog Component

sophistifunk:
**Description**
- See [JENKINS-39371](https://issues.jenkins-ci.org/browse/JENKINS-39371).

**Submitter checklist**
- [X ] Link to JIRA ticket in description, if appropriate.
- [X ] Change is code complete and matches issue description

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

Squashed commit of the following:

    josh/JENKINS-39371-dialog-2 * Fix some remaining TODOs and an issue with proptypes
    josh/JENKINS-39371-dialog-2 * Fix some flow / lint issues
    josh/JENKINS-39371-dialog-2 * Add wrapper component to do it all easily, add some styles for input / error etc
    josh/JENKINS-39371-dialog-2 * Add header component
    josh/JENKINS-39371-dialog-2 * Fix a small flow issue
    josh/JENKINS-39371-dialog-2 * Wire it up
    josh/JENKINS-39371-dialog-2 * Initial merge of Dialog from old branch (won't run yet)

### DIFF
--- a/less/components/popups.less
+++ b/less/components/popups.less
@@ -1,5 +1,3 @@
-// TODO: Rename this file because it's for assorted modal popups not just popover
-
 //--------------------------------------------------------------------------
 //
 //  Popups, popovers, and dialogs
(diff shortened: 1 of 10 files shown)


i386:
FYI @kzantow this is the dialog for your editor step listing stories.

less/components/popups.less
```diff
@@ -69,12 +67,54 @@
     align-items: center;
     width: 100%;
     height: 80px;
-    min-height: 80px;
+    flex-shrink: 0;
```
kzantow: can we use flex for this? I thought we had to change away from flex to get IE working properly

sophistifunk: This is news to me. If we see bugs on whatever is the minimum version of IE we support I guess we can address it there, but I'd go with an ie-specific stylesheet rather than dropping flex-box, as it's miles better and it's *everywhere* in the editor.

cliffmeyers:
Been bogged down in a few other things today but this will be priority one tomorrow, if not before.

less/components/popups.less
cliffmeyers: I think the problems with flexbox were limited to how the top-level page template in `blueocean-web` was laid out, causing IE issues. Perhaps @tfennelly can clarify since I think he was the one who fixed them.

src/js/components/dialog/Dialog.jsx
```diff
@@ -0,0 +1,212 @@
...
+/** Basic header for dialogs */
+export class DialogHeader extends Component {
+    props: {
+        children?: ReactChildren
+    };
+
+    render() {
+        return (
+            <div className="Dialog-header">
+                <h3>{this.props.children}</h3>
```
cliffmeyers: In the event someone wanted to use `Dialog` but provide custom header content (icons, a button, etc), should we do anything a bit fancier here? Perhaps check if `children` is a string and wrap in h3, if a `ReactElement` omit the h3?

I know they can always use `BasicDialog` and built up their own, just wasn't sure it was a use-case we wanted to bother supporting.

PR #114, opened 2016-11-14
Task/jenkins 39563 basic form comps

cliffmeyers:
**Description**
- See [JENKINS-39563](https://issues.jenkins-ci.org/browse/JENKINS-39563).
- This is a first pass at a couple of simple controls and some plumbing for forms and display validation errors. I thought it might be good to adopt of pattern of building basic controls (`TextInput`, `RadioButtonGroup`) that are simple, have minimal DOM clutter and are easy to compose. Then I created the `FormElement` wrapper that we can use for basic layout and error message presentation. Then I wrapped up `TextInput` and `FormElement` into a `FormTextInput` (not sure I like the name) with the idea that we could support common validation use cases via props, like "required" or "validatorRegex" or "validatorFunc".  

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
cc @sophistifunk for feedback on the integration w/ FormElement and such. I want to nail down that API and make sure we're comfortable before I start building out other components and skinning everything up.

cliffmeyers:
I added a `Checkbox` component and also took a stab at how we might handle supporting the small, medium and large sizes in the style guide: 1. classes / mixins for each size of a component, typically not used directly unless there's a need to override in a one-off case, 2. applying those mixins to each component when they are nested within a container with a `layout-small|medium|large` class/mixin. We could apply those layouts as mixins to more semantic container names if we can come up with ones that make sense.

cliffmeyers:
Refactored `Favorite` to just be a custom `Checkbox`.

sophistifunk:
Just starting to look into this now. FWIW, I've found this to be a good starting point for organising component styles, and I've tried to stick pretty close to it for things I've built recently: https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md

It seems pretty sensible on the whole, but it might have some blurry edges when we're creating components to be used in the nebulous world of future, third part code :)

less/components/checkbox.less
```diff
@@ -0,0 +1,77 @@
...
+        border: 0;
+        clip: rect(0, 0, 0, 0);
+        height: 1px;
+        margin: -1px;
+        overflow: hidden;
+        padding: 0;
+        position: absolute;
+        width: 1px;
+    }
+
```
sophistifunk: How does this work with keyboard tab-focus?

cliffmeyers: It does work. That's really the reason for those ugly combination of flip / height / margin junk - so you can skin the other element but let the browser still recognize checkbox by not hiding it entirely.

less/components/radio-button-group.less
```diff
@@ -0,0 +1,25 @@
+.radio-button-group {
+
+    .radio-button-wrapper {
```
sophistifunk: Why not just put these -wrapper styles on the parent div and ditch the extra wrapper, since there's no styles on the outer `.radio-button-group`?

src/js/components/favorite/Favorite.jsx
```diff
@@ -1,73 +1,40 @@
...
+                className={`Favorite ${extraClass}`}
+                checked={this.props.checked}
+                onToggle={this.props.onToggle}
+                label={this.props.label}
             >
-                <input type="checkbox"
-                       onChange={this.toggle.bind(this)}
-                       checked={this.state.checked} />
-
                 <svg className="star-icon" width="288" height="24" viewBox="0 0 288 24" xmlns="http://www.w3.org/2000/svg">
```
sophistifunk: We should probably extract complex SVG shapes (more than a handful of lines) into separate components IMO, it makes them easier to use, and makes the `render()` of this compound component easier to grok. Doesn't need to be a fully polished exported component until we need it, but I find even a file-local component makes it easier to read.

cliffmeyers: So define a stateless function component `FavoriteStar` within the file and then drop it Favorite's render method? Sure, I can absolutely do that.

sophistifunk:
Looks good, minor nitpicks aside 😁 

bee

less/sizing.less
```diff
@@ -0,0 +1,22 @@
+
+.layout-small {
+    font-size: 12px;
+
+    .Checkbox {
```
cliffmeyers: Right on, totally forgot about that syntax. You like the convention of `layout-small|medium|large` then I take it?

sophistifunk: Seems OK, we can try it on for size at least :)

cliffmeyers:
@sophistifunk did you have any feedback on the `FormElement` and the concept of wrapping up a `TextInput` + `FormElement` into a more feature-rich control? If not, that's cool, I'll keep moving in that direction.

sophistifunk:
I like having a wrapping component, but I'd like to see the `<label for="fooBar"/>...<input id="fooBar"/>` pairing. But this can complicate checkboxes / radios, which have their own label (which should have preference in `for=""`) as well.

less/components/radio-button-group.less
cliffmeyers: Fixed in commit 5a7eafd

src/js/components/favorite/Favorite.jsx
cliffmeyers: Fixed in dd8d21d

less/sizing.less
cliffmeyers: Fixed in commit 2765c52

cliffmeyers:
@sophistifunk I have taken this about as far as I would like in this PR, otherwise it is just going to become an unwieldy monster :) Perhaps we can discuss the `label` semantics when we have a few mins? There are some considerations (like ID collisions) for `label for=''` - maybe we want to evaluate the trade-offs. I had thought a long time ago I confirmed that nesting content in `label` is legit per W3C spec but will need to dig to look.

sophistifunk:
Works for me mate.

cliffmeyers:
@sophistifunk sweet. Maybe you want to have a peek at the new commits - might be the easiest way for you to cut through the clutter. Thanks!

sophistifunk:
Had a quick look, it all looks sensible to me. The :bee stands :)

PR #115, opened 2016-11-17
Add close button to dialog

kzantow:
**Description**
Add a standard close button to dialog:

image

**Submitter checklist**
- [x] Change is code complete

@i386 @sophistifunk @michaelneale  PTAL, not sure you want this but I think it makes sense to have a standard close button on dialogs, could be hidden by default, up to you...

### DIFF
--- a/less/components/popups.less
+++ b/less/components/popups.less
@@ -73,21 +73,53 @@
     border-top-left-radius: @popup-border-radius;
     border-top-right-radius: @popup-border-radius;
     padding: 0 32px;
+    position: relative;
+
+    .Dialog-close-button {
+        position: absolute;
+        right: 21px;
+        top: 50%;
+        font-size: 36px;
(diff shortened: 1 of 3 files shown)


sophistifunk:
This will look great with our "giant close button used to indicate failure" feature. :-|

kzantow:
@sophistifunk hahaha... ehh.... maybe there are some ideas to change the 'giant close button to indicate failure'

i386:
@brody ping

kzantow:
@brody WDYT about a standard close button like this?

brody:
@kzantow The Primary button should be whatever the action is (eg Submit) and the secondary button should be "Close" or "Cancel". Hitting ESC or clicking outside the modal would also dismiss it.
Let me know if you can think of any edgecases where the above pattern wouldn't work.

image

PR #118, opened 2016-11-22
Bug/general table issues

sophistifunk:
**Description**

This should serve as a base on which to cook up a nicer solution to the problem of giving table rows middle-clickable `<a href="">`s

* Consolidate table styles
* Replace CSS tricks for generating header/body divide with an explicit hr row
* Split out the padding into a specific classname, and add a generic classname for row highlight
* Better test data in table stories
* Fix some sources and tweak rules to squash all lint warnings/errors

- See [JENKINS-39927](https://issues.jenkins-ci.org/browse/JENKINS-39927).
- See [JENKINS-39853](https://issues.jenkins-ci.org/browse/JENKINS-39853).

**Submitter checklist**
- [X ] Link to JIRA ticket in description, if appropriate.

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


sophistifunk:
Ping @imeredith @cliffmeyers

src/js/stories/TableStories.jsx
```diff
@@ -0,0 +1,122 @@
...
+const detailedData = [
+    [ 'Ryu',      '21-7-1964',  '5\'10"', '150 lbs', 'O' ],
+    [ 'E. Honda', '3-11-1960',  '6\'2"',  '304 lbs', 'A' ],
+    [ 'Blanka',   '12-2-1966',  '6\'5"',  '218 lbs', 'B' ],
+    [ 'Guile',    '23-12-1960', '6\'1"',  '191 lbs', 'O' ],
+    [ 'Ken',      '14-2-1966',  '5\'10"', '169 lbs', 'B' ],
+    [ 'Chun-Li',  '1-3-1968',   '5\'8"',  '-',       'A' ],
+    [ 'Zangief',  '1-6-1956',   '7\'',    '256 lbs', 'A' ],
+    [ 'Dhalsim',  '22-11-1962', '5\'10"', '107 lbs', 'O' ]
+];
```
cliffmeyers: 👍

src/js/components/Table.jsx
```diff
@@ -1,71 +1,111 @@
...
+            disableDefaultPadding } = this.props;
 
-    getLabel(column) {
-        if (typeof column === 'string') {
-            return column;
-        }
-        return column.label;
-    }
+        const divider = headers && headers.length && !disableHeaderDivider ?
+            <TableDivider numCols={headers.length}/> : '';
```
cliffmeyers: We should return `null` rather than empty string to avoid this warning:

`Warning: validateDOMNesting(...): Whitespace text nodes cannot appear as a child of <thead>. Make sure you don't have any extra whitespace between tags on each line of your source code. See Table > thead > #text.`

src/js/stories/TableStories.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+function renderRow(values) {
+    return (
+        <tr>
+            { values.map(value => <td>{value}</td>) }
```
cliffmeyers: Perhaps add `key` attribute to the `td` (also possibly the `tr`) to avoid the https://fb.me/react-warning-keys warning in console.

sophistifunk: Good catch, I was wondering where that came from.

src/js/components/Table.jsx
sophistifunk: Odd. Both in that it's wrong, sort of, and that I thought React collapsed them all down to nothing when there's no actual text nodes. I'll change it anyway :)

sophistifunk:
I'm going to turn off the no-wrap as default behaviour, because it doesn't work without fixed layout, which is off by default. It seems better to remove unusual default behaviour rather than add more for consistency.

cliffmeyers:
@sophistifunk if you could please merge and publish this today, I can get it integrated into jenkinsci/blueocean-plugin#614 into and finish that one up. Will put in a little time on Fri to get it ready for final review / merge.

PR #119, opened 2016-11-22
Bug/dropdown bugs

cliffmeyers:
**Description**
- Just fixing a few small bugs w/ the dropdown

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
Pretty simple... but perhaps @sophistifunk wants to have a look!

sophistifunk:
Looks good. 🐝  

What about having an additional className when we show a placeholder label instead of a selected item label, so it can be in italics, or half-opacity or something to show it's not a real value? Or does this feature already exist and I just missed it because this PR doesn't touch that code? :)

cliffmeyers:
It does not exist, although interestingly the style guide doesn't any special text formatting for the placeholder. I'm fine adding that class though, will do so here shortly.

cliffmeyers:
@sophistifunk another batch of changes here... 

- Added class when placeholder is visible so it can be styled
- Fixed a bug where in Safari and Edge the menu would flicker in top-left corner when first opening. I added a config option to FloatingElement that when enabled will position it immediately after rendering rather than waiting for the animation frame.
- Fixed a bug where dropdown with a longer placeholder / label was overflowing its parent. This was actually quite tricky to get right and doesn't work in IE11 - but I think it's an edge case and going to wait until it actually surfaces an issue before I spend more time.
- Fixed a bug where focus wasn't always given to selected item when opening the menu.

src/js/components/FloatingElement.jsx
```diff
@@ -422,12 +422,16 @@ export class FloatingElement extends Component {
     }
 
     componentRendered() {
-        // As soon as possible, we need to re-calculate our position
-        this.validatePositioningScheduled = true;
-        window.requestAnimationFrame(() => {
-            this.validatePositioningScheduled = false;
+        if (this.props.positionImmediate) {
```
sophistifunk: This is OK for now but I think we should re-visit it in the future, perhaps have it only turned on based on detected browser, or perhaps set the popup's style.opacity to 0 initially, then set it to 1 once the position has been set.

Actually, setting the initial values for position to something off-screen may make the problem go away entirely, I haven't yet got a Win10 VM set up to test it.

src/js/stories/DropdownStories.jsx
```diff
@@ -13,48 +15,114 @@ const style = {
...
+        const label = `${text} ${options.length + 1}`;
+        options.push(!asObject ? label : { label });
     }
     return options;
 }
 
+
+
 function Default() {
+    Utils.createCssRule(
```
sophistifunk: This is an interesting approach, but I wonder how often Storybook reloads the entire iFrame, or if we might end up creating a whole bunch of extra `<style/>` nodes as you poke around trying out different stories.

src/js/components/FloatingElement.jsx
cliffmeyers: Not opposed to that at all. We need to double-check the opacity hack; I seem to recall Safari doing something really weird like giving keyboard focus to the elements in the menu even though opacity is 0; that's why I had to switch to creating/destroying the DOM when the menu opened and closed.

sophistifunk: I didn't mean keep it around all the time, just when you need it create it with opacity 0, then once it's in the right place, "one frame" set it to 1, so you don't see it appear for a "frame" at 0,0 then blink into the right place.

Or, we just set the initial position as -9999, -9999, see how that goes.

I'm just throwing out ideas to fix the original problem without having to block everything for a synchronous dom validation

src/js/stories/DropdownStories.jsx
cliffmeyers: You're right, it can create some dup style rules and be a bit funky. I could defend against that within the method? I guess my thought was for storybook it was ok to do things a little hacky, but I don't mind trying to clean it up before we merge.

michaelneale:
status of this?

src/js/components/FloatingElement.jsx
cliffmeyers: Fixed in commit 9378d0c

src/js/stories/DropdownStories.jsx
cliffmeyers: Fixed in f7fda7e. I cleaned up this method a bit to avoid creating duplicate rules which happens by default (presumably due to some webpack loading oddities). Only catch is that if you remove a rule you need to do a full page refresh to get rid of it; otherwise it seems to work pretty well in all browsers.

cliffmeyers:
@michaelneale I think this is ready to merge, just giving @sophistifunk a chance for any final feedback... but I think we are good.

sophistifunk:
bbm

PR #127, opened 2016-12-30
Task/form element enhancements

cliffmeyers:
**Description**
- Controls wrapped in `FormElement` will get some sensible vertical spacing by default, scaled based on a parent container with `className=layout-small|medium|large`
- Optional divider below title
- Error states

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
cc @sophistifunk @kzantow @scherler 

Couple discussion points, very interested in team's feedback:
- Which `display` value makes sense for controls by default? Seems like `inline-block` is used in most `input` controls and some popular custom controls. Would `inline-block` or `inline-flex` be reasonable?
- For `FormElement` it seems `block` or `flex` is reasonable so that common form layouts just stack vertically like you'd expect.
- I'm seeing this come up because a `FormElement` using a divider with a nested control doesn't necessarily display the way I might expect by default. Then again, maybe it's doing the right thing when placed in a full-width container?

cliffmeyers:
form-element-1-basics
form-element-2-sizing
form-element-3-sizing-dividers
form-element-4-errors

cliffmeyers:
bump... cc @sophistifunk @kzantow @scherler

sophistifunk:
Is there a use-case for the title / element divider?

sophistifunk:
Code looks good BTW, 🐝 As for `display:` I think inline-block or inline-flex will be what most individual "widgets" are likely to need, but I definitely think we want flex for all wrappers or "groups of" components. For example, a checkbox and its label would be in an inline-flex by default, but a group of them should be flex, as you're not likely going to be embedding it alongside body text very often, and if they're parented by a `display:flex` container rather than standard flow layout, the container's rules will mostly determine the layout anyway.

PR #134, opened 2017-01-20
Task/jenkins 38598 list component

cliffmeyers:
**Description**
- Building a reusable List control.

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/src/js/components/List.jsx
+++ b/src/js/components/List.jsx
@@ -1,5 +1,6 @@
 import React, { PropTypes } from 'react';
 import KeyCodes from '../KeyCodes';
+import Utils from '../Utils';
 
 type Props = {
     className?: string,
@@ -46,6 +47,8 @@ export class List extends React.Component {
         this.state = {
             selectedItem: null,
         };
+
+        this.groupId = Utils.randomId('List');
     }
 
     componentWillReceiveProps(nextProps:Props) {
@@ -84,12 +87,8 @@ export class List extends React.Component {
         this._selectItem(index, item);
     }
 
-    _onKeyPressListItem(event:React$Event, index:number, item:Object) {
-        event.preventDefault();
-
-        if (event.nativeEvent.keyCode === KeyCodes.SPACEBAR) {
-            this._selectItem(index, item);
-        }
+    _onChangeSelection(index:number, item:Object) {
+        this._selectItem(index, item);
     }
 
     _selectItem(index:number, item:Object) {
@@ -120,13 +119,7 @@ export class List extends React.Component {
 
         const listClass = this.props.className || '';
         const selectedClass = this.state.selectedItem ? 'List-selected' : '';
-
-        let containerClass = '', itemClass = '';
-
-        if (this.props.defaultStyles) {
-            containerClass = this.props.defaultContainerClass ? 'u-default-list-container' : '';
-            itemClass = this.props.defaultItemClass ? 'u-default-list-item' : '';
-        }
+        const containerClass = this.props.defaultStyles ? 'u-default-list-container' : '';
 
         const labelFunc = labelFunction || itemToLabel;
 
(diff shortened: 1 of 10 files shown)


cliffmeyers:
@sophistifunk i need to resolve a few Flow errors on this still - if you have any spare time maybe you could pull down and have a look? Otherwise I think this about ready for an initial go. Screenies to come shortly.

cliffmeyers:
list-1-default-list-and-hover

list-2-selected-item-and-squished

cliffmeyers:
cc @sophistifunk

sophistifunk:
LGTM

bee-movie-bee

src/js/components/List.jsx
```diff
@@ -0,0 +1,197 @@
...
+    _onClickListItem(event:Event, index:number, item:Object) {
+        event.preventDefault();
+
+        this._selectItem(index, item);
+    }
+
+    _onKeyPressListItem(event:React$Event, index:number, item:Object) {
+        event.preventDefault();
+
+        if (event.nativeEvent.keyCode === KeyCodes.SPACEBAR) {
```
kzantow: So, I got a bit lazy on my list implementation before. I think the keyboard interaction with this should be a bit different, basically like that of radio buttons, tab to the list, use arrow keys to navigate the items within it. Also, not sure the default behavior should be to select an item in the list but rather just call a callback and let the caller decide what happens. For example, I already had to change the behavior of my list to click/enter/etc. select and dismiss the containing dialog.

src/js/components/List.jsx
```diff
@@ -0,0 +1,197 @@
...
+            console.error(`'children' supplied to List must be zero or one elements`);
+            return null;
+        }
+
+        const listClass = this.props.className || '';
+        const selectedClass = this.state.selectedItem ? 'List-selected' : '';
+
+        let containerClass = '', itemClass = '';
+
+        if (this.props.defaultStyles) {
```
kzantow: Not sure I agree with the default styling here. I'd rather have a functional shell of a list that could have a top-level class applied and style it in some different, common ways.

kzantow:
I'd rather see this component a bit more generic with better keyboard handling, perhaps with some specific implementation to apply these default styles, like a `StyledList`, which really just adds a top-level class... WDYT @cliffmeyers ?

src/js/components/List.jsx
scherler: @kzantow callback :+1:

cliffmeyers: re: keyboard I agree, having to tab through every item is pretty poor. I might actually use "hidden" radio buttons to get the right accessibility, since OS X has some significant default behavior for how keyboard works and I want to respect that without significant custom code.

re: selection, it seems a reasonable default behavior to me. If we discard it, are you suggesting that "selectedItem" becomes a prop that the developer needs to populate? I'm a little torn on that since the overwhelming use case for Lists is to allow something to be selected.

cliffmeyers:
@kzantow I agree that the styling can be simplified so the JDL applies the default styling when that special "default" class is applied. I do think for visual consistency that these components should "do the right thing" by default, and by that I mean they should conform to the style guide. So perhaps we stick with a 'defaultStyles' prop that - when active - applies a top level class that applies the proper styling for the container and the list items. And the "DefaultRenderer" will apply its default styling as it does today. If a developer wants full control, they just flip `defaultStyles` to false and do whatever they like.

The reservation I have about `List` vs `StyledList` is that I also need to build a `FilterableList` component that composes `List`. So will we have `List`, `StyledList`, `FilterableList` and `StyledFilterableList`? That feels like a slippery slope too a whole lot of components that really just differ based on a boolean prop. I totally get your concern about avoiding "god components" - and agree - but I don't think splitting components up based on the their styles is the best way to go.

src/js/components/List.jsx
scherler: not sure about that.

I do not see any evil in the styles but I would have done it like:

```
const { containerClass = 'u-default-list-container', itemClass = 'u-default-list-item' } = this.props.defaultStyles;
```

So you have either the default style or the one you pass as default. This way it is easy to apply styles from the caller.

src/js/components/List.jsx
cliffmeyers: Keyboard accessibility improved in 06cccb4

cliffmeyers:
I tried to simplify down some of the styles in 06cccb4. By default they are applied but the developer can opt-out with `defaultStyles=false` and if desired use a custom `className` to style it however they like.

cliffmeyers:
Incorporated feedback where appropriate. I'm going to leave the `onItemSelect" callback as-is for reasons explained above.

PR #135, opened 2017-01-23
[FIX JENKINS-39857_Display_steps_durations_in_Blue_Ocean]

scherler:
… not use humanize() format but a custom one

**Description**
- See [JENKINS-39857](https://issues.jenkins-ci.org/browse/JENKINS-39857) and https://issues.jenkins-ci.org/browse/JENKINS-41282.

Allow that ResultItem to accept an object as well (not only a string as before). Allow timeDuration to not use humanize() format but a custom one.

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@jenkins-cd/design-language",
   "jdlName": "jenkins-design-language",
-  "version": "0.0.105-unpublishedTime",
+  "version": "0.0.105-TimePrecise",
   "description": "Styles, assets, and React classes for Jenkins Design Language",
   "main": "dist/js/components/index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@jenkins-cd/design-language",
   "jdlName": "jenkins-design-language",
-  "version": "0.0.105-TimePrecise",
+  "version": "0.0.106-unpublished",
   "description": "Styles, assets, and React classes for Jenkins Design Language",
   "main": "dist/js/components/index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@jenkins-cd/design-language",
   "jdlName": "jenkins-design-language",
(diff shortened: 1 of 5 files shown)


i386:
@scherler What's wrong with the moment format we use everywhere right now?

scherler:
@i386 nothing, we only need to allow that we can use a custom format as requested in the ticket

scherler:
https://github.com/jenkinsci/blueocean-plugin/pull/743 is the user

i386:
@scherler oh I think the existing moment format is fine - the mockup may be a bit out of date

scherler:
@i386 it does not hurt neither. liberty to the people ;)

i386:
@scherler I think a user defined that mockup - not sure what happened to our one

michaelneale:
this should be similar to the header - but I think the header says "a few seconds" - I am pretty sure there was tickets I may have accidentally closed that assumed the header was different. 

The user did not request it be customisable, they just suggested a short hand display, but whatever it is, should be the same as the header right @i386? Why not make it consistent

i386:
@michaelneale actually there is a ticket in the backlog to adjust it from "A few seconds" to "X seconds". Yes, consistency is key to making our UI great :)

i386:
@scherler if you are in the area https://issues.jenkins-ci.org/browse/JENKINS-41282 changes "A few seconds" to "X seconds".

michaelneale:
right, so if the header is updated as per that ticket, then the steps can use the same logic right? (and it will be closer to what Julian wanted). Win win!

Nice one @scherler !

i386:
@michaelneale thats right!

michaelneale:
yeah unfortunately, I think https://issues.jenkins-ci.org/browse/JENKINS-41282 should be done at the same time (combine the two) so that the same code can be used in both.

scherler:
@i386 and @michaelneale https://issues.jenkins-ci.org/browse/JENKINS-41282 is now part of this fix

i386:
Nice one :)

PR #139, opened 2017-02-02
Hotfix/icon button ffx

cliffmeyers:
**Description**
- See commit comments

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/less/components/icon-button.less
+++ b/less/components/icon-button.less
@@ -1,7 +1,10 @@
 button.IconButton {
     .IconButton-medium;
-    display: inline-flex;
-    align-items: center;
+
+    .IconButton-wrapper {
+        display: flex;
+        align-items: center;
+    }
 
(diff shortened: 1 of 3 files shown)


cliffmeyers:
cc @sophistifunk as you have a good memory for the font history. I could have sworn that LatoLatinWeb was being used across the board. It could be that when I restyled the buttons that a `font-family` was no longer applied directly to button, allowing the normalize reset to `sans-serif` to take over. So I decided it was approach to negate that on the common form controls. Thoughts?

sophistifunk:
Uh, there was a problem with IE, I think. Maybe something @imeredith fixed.

cliffmeyers:
Discussed with @imeredith in chat and he was not aware of any issues. I will merge this PR shortly.

PR #142, opened 2017-02-03
Bug/jenkins 38598 list fixes

cliffmeyers:
**Description**
- Resolve height / overflow issues by simplifying dom
- Fix a bug where the focused item would not always be visible in the list (due to scrolling)
- Fix a bug where the defaultSelection item did not initial focus when tabbing into control
- Refine behavior so that "selectedItem" is discard if data changes and is null or doesn't contain "selectedItem"
- Add support for disabling the list
- Fix a few flow errors

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/src/js/components/List.jsx
+++ b/src/js/components/List.jsx
@@ -166,7 +166,7 @@ export class List extends React.Component {
     }
 
     render() {
-        const { children, data, keyFunction, labelFunction } = this.props;
+        const { children, data, disabled, keyFunction, labelFunction } = this.props;
 
         const childCount = React.Children.count(children);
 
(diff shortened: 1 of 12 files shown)


cliffmeyers:
cc @sophistifunk

src/js/components/List.jsx
```diff
@@ -7,6 +9,7 @@ type Props = {
     style: Object,
     data?: Array<Object>,
     labelFunction?: Function,
+    disabled?: bool,
```
sophistifunk: Not saying to change this because it's fine, just a tip that you might get less curry from Flow if instead of making things like disabled optional, you just add a false / empty value to `defaultProps`

sophistifunk:
Looks good 🐝

src/js/stories/ListStories.jsx
```diff
@@ -95,26 +89,114 @@ function RendererOptions() {
...
+        justifyContent: 'space-around',
+        alignItems: 'center',
+        padding: 10,
+        height: 400,
+    };
 
     return (
-        <div style={{...style, height: 300}}>
-            <p>This Layout is useful for demonstrating keyboard accessibility and focus behavior.</p>
+        <div style={outer}>
```
scherler: Why not use `class` attribute instead of `style`

cliffmeyers: We haven't really been using classes in any of the storybooks but we could at some point if we wanted to make the design of them more consistent. Inline styles are sorta convenient with the webpack hot reloading...

sophistifunk: There aren't many times we should have custom stylesheets for stories, inline styles are definitely the way to go for almost every use.

PR #148, opened 2017-02-16
Task/jenkins 42109 text input icons

cliffmeyers:
**Description**
- See [JENKINS-42109](https://issues.jenkins-ci.org/browse/JENKINS-42109).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
cc @sophistifunk I actually went in a different direction from our last chat. Explanation:

I suspect we're going to need a prop like `clearButton` which gives the user a little "X" on the right after they type into it, which can click to clear the text. I haven't seen it in the comps (yet) but it seems like a pretty obvious thing to have.

So, if icons are only supported by a separate `IconTextInput` then it seems odd to require a developer to use `IconTextInput` just to get that "clear" button... the impl details almost kinda leak up to the API at that point. 

Now, maybe it's just a naming thing... and maybe `AdvancedTextInput` should support these extra fancy features while `TextInput` remains very barebones. That said, the code to get this working was pretty darn simple, so it seem fair to leave in there. At any rate, the refactoring here was worthwhile (especially the changes to `TextControl`) so if we decide we really should split out `ATI` then I think I can do that without too much pain.

I left this as WIP only because I need to take one more pass to support our small / medium / large text sizes better. Will refactor that evil margin-left/right CSS into a mixin so it reads better.

cliffmeyers:
@sophistifunk perhaps you can review this one when you are online today? Thanks :)

less/components/text-input.less
```diff
@@ -18,23 +29,77 @@
...
     }
+
+    .TextInput-icon {
+        fill: #999;
+        cursor: default;
+    }
 }
 
+
+.icon-rules(@icon-size, @icon-padding) {
```
sophistifunk: 👍

sophistifunk:
This is less code than I was expecting, nice wun bwuvva. 🐝

src/js/components/forms/TextControl.js
```diff
@@ -53,16 +53,24 @@ export class TextControl extends React.Component {
...
-                        onChange: e => this._onChange(e),
-                        onBlur: e => this._onBlur(e),
+                { React.Children.map(this.props.children, child => {
+                    // while multiple children can be passed in (icons, etc)
+                    // we only want to pass down props to underlying text control
+                    if (child && (child.type === 'input' || child.type === 'textarea')) {
+                        return React.cloneElement(
+                            child,
+                            {
+                                placeholder: this.props.placeholder,
```
kzantow: Might be useful to pass `this.props.inputProps` or something along for all the other stuff that might be needed, e.g. `onFocus` etc.

src/js/components/forms/TextInput.jsx
```diff
@@ -1,12 +1,35 @@
...
+NestedIcon.propTypes = {
+    className: PropTypes.string,
+    icon: PropTypes.string,
+};
+
+
 export class TextInput extends React.Component {
 
     render() {
+        const classLeft = this.props.iconLeft ? 'u-icon-left' : '';
```
kzantow: what if you need multiple icons? (already had the column filter design needing that... well, really just needing extra elements)

src/js/components/forms/TextInput.jsx
```diff
@@ -1,12 +1,35 @@
...
+        const classRight = this.props.iconRight ? 'u-icon-right': '';
+
         return (
-            <TextControl {...this.props} className={`TextInput ${this.props.className}`}>
+            <TextControl {...this.props} className={`TextInput ${this.props.className} ${classLeft} ${classRight}`}>
+                { classLeft && <NestedIcon className={classLeft} icon={this.props.iconLeft} /> }
```
kzantow: what happens if you want these icons to be clickable?

cliffmeyers: I'll certainly add that in once we have a use case for it. Right now I'm supporting "icons as decoration" only.

src/js/components/forms/TextInput.jsx
cliffmeyers: Can you point me to the design for that? You're saying there are multiple icons on the left or right?

cliffmeyers:
@kzantow your comments here are valid, although I think they're all around future use cases and we can refine them as we go. Need to get this one through as it is blocking #151 and jenkinsci/blueocean-plugin/845. I will make sure this good feedback is taken into account when the time comes.

PR #150, opened 2017-02-20
Clean up some old components / styles no longer needed

sophistifunk:
Clean up some old components / styles no longer needed

@reviewbybees


sophistifunk:
This is just clean-up of some unused components and a few css classes that we're no longer using.

michaelneale:
I love the smell of deleted code in the morning, :bee: as long as you are comfortable (perhaps want to get @imeredith to double check as well, as not easy to test that none are needed)

cliffmeyers:
LGTM 🐝  No surprises here as far as what was deleted. Good deal.

PR #153, opened 2017-02-23
Hotfix/text input icon tweaks

cliffmeyers:
**Description**
- Icon layout had some spacing inconsistencies that were masked by setting a min-width, and min-width was problematic for some views in the editor. instead use absolute positioning which is simpler and requires far fewer compensating margin "hacks". The downside is that we don't get default vertical centering of icons from flexbox, but it doesn't really matter since we are setting icon sizes explicitly anyways.

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/less/components/text-input.less
+++ b/less/components/text-input.less
@@ -90,7 +90,7 @@ input[type="text"] {
 }
 
 .TextInput-medium {
-    .icon-rules(24px, 6px, 10px);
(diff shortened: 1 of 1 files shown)


cliffmeyers:
@kzantow yeah... as neat as flexbox is, it was not appropriate to use here :P

PR #154, opened 2017-02-27
Update react-storybook to latest version to fix iframe on docs site

halkeye:
**Description**
- Currently https://jenkinsci.github.io/jenkins-design-language/docs iframes the storybook pages. Unfortuatly due to an existing bug, the iframes don't work
- See https://github.com/storybooks/react-storybook/issues/555
- This has already been addressed in later versions of react-storybook, so this PR just updates the dependancy

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "react-dom": "15.3.2"
   },
   "devDependencies": {
-    "@kadira/storybook": "2.24.1",
+    "@kadira/storybook": "^2.35.3",
     "babel-eslint": "7.0.0",
     "babel-polyfill": "6.16.0",
     "babel-preset-es2015": "6.16.0",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "react-dom": "15.3.2"
   },
   "devDependencies": {
-    "@kadira/storybook": "^2.35.3",
+    "@kadira/storybook": "2.35.3",
     "babel-eslint": "7.0.0",
     "babel-polyfill": "6.16.0",
     "babel-preset-es2015": "6.16.0",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -14,9 +14,9 @@
       "dev": true
     },
     "@kadira/storybook": {
-      "version": "2.24.1",
-      "from": "@kadira/storybook@2.24.1",
-      "resolved": "https://registry.npmjs.org/@kadira/storybook/-/storybook-2.24.1.tgz",
+      "version": "2.35.3",
+      "from": "@kadira/storybook@latest",
+      "resolved": "https://registry.npmjs.org/@kadira/storybook/-/storybook-2.35.3.tgz",
       "dev": true
     },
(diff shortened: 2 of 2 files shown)


michaelneale:
cc @sophistifunk @cliffmeyers look ok?

sophistifunk:
Does this build in our cloud? Github is telling me it failed, but the URL is bad.

michaelneale:
@sophistifunk yeah it does - but a test instance (noting to do with dogfood) wallopped the status. 

@dragoonis if you push a change it shoudl trigger another build.

michaelneale:
@dragoonis you have an invitation to be a collaborator on this repo so you can work in branches here too if that suits.

cliffmeyers:
@michaelneale this PR doesn't show up on the PR's tab on Dogfood. Should it? Otherwise I'm not sure how to re-run the build.

michaelneale:
I don't know either - convert it to a branch and it should rebuild.

halkeye:
I think https://github.com/jenkinsci/jenkins-design-language/blob/master/site-publish.sh isn't part of the build anymore.

sophistifunk:
> I think https://github.com/jenkinsci/jenkins-design-language/blob/master/site-publish.sh isn't part of the build anymore.

Never was, as it doesn't work for everybody.

scherler:
@sophistifunk what is the problem. I fixed it around a month ago when i published the last time

sophistifunk:
IIRC, it required a specific version of git that's newer than what you get on OSX.

scherler:
@sophistifunk yeah I added:

```
git merge --no-edit -s ours remotes/webpages/gh-pages --allow-unrelated-histories
```

` --allow-unrelated-histories` is to allow that merge if you are on git 2.9+

package.json
```diff
@@ -47,7 +47,7 @@
     "react-dom": "15.3.2"
   },
   "devDependencies": {
-    "@kadira/storybook": "2.24.1",
+    "@kadira/storybook": "^2.35.3",
```
scherler: `"@kadira/storybook": "2.35.3"` please do not use `^`

halkeye:
Doesn't yarn.lock handle that anyways? I can't fix this or on mobile so I'll try and get to it asap

package.json
cliffmeyers: I was thinking, should we write a utility that validates package.json for semver best practices. My feeling is that deps and devDeps should use exact, but peerDeps should use caret operator. We could create a standalone util for use in JDL but also bake it into js-builder by default. Or maybe something comparable already exists, can look. WDYT @scherler ?

michaelneale:
@halkeye yarn isn't in use yet (and probably better to be explicit, like you have been). 

@scherler package.json has been updated.

michaelneale:
ping @scherler @sophistifunk can we merge this one in now?

sophistifunk:
Definitely

michaelneale:
nice one @halkeye !

*swish*

PR #155, opened 2017-02-28
New table components

sophistifunk:
**Description**

**Submitter checklist**
- [X] Link to JIRA ticket in description, if appropriate.

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


sophistifunk:
There's going to be rough edges I discover in this as I apply it to various BO screens, but I want to get it up so that I can recruit people to look over the code, and get some feedback about the visuals via storybook.

michaelneale:
Looking good!

In JTable/Manual - I can't mouse over the second last row for some reason (no highlight appears) - which seems... odd - expected?

i386:
@michaelneale @sophistifunk I also was able to reproduce the second last row not highlighting on hover.

i386:
💯  looks great

michaelneale:
I could have used the word "penultimate" but it's too late now.

src/js/components/jtable/TableCell.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const classNames = ['JTable-header'];
+
+    if (className) {
+        classNames.push(className);
+    }
+
+    const newProps = {
+        ...props,
+        children: undefined,
```
cliffmeyers: Does `children` need to be explicitly set to `undefined` like this?

src/js/components/jtable/TableRow.jsx
```diff
@@ -0,0 +1,145 @@
...
+
+        if (typeof href === 'string' && href.length > 0) {
+            // We switch to an <A> instead of <DIV> so the user can middle-click
+            tagName = 'a';
+            classNames.push('JTable-row--href');
+        }
+
+        props.className = classNames.join(' ');
+
+        return React.createElement(tagName, props, ...newChildren);
```
cliffmeyers: Should we make `href` and `onClick` mutually exclusive properties? Or perhaps we just leave it up to the `onClick` handler to do `preventDefault()` ?

src/js/components/jtable/JTable.jsx
```diff
@@ -0,0 +1,137 @@
...
+ * over layout, and allows us to do things like making a whole row an anchor, etc.
+ */
+export class JTable extends Component {
+
+    /**
+     * Just a simple helper
+     */
+    static column(width: number, name: string = '', isFlexible: boolean = false): ColumnDescription {
+        return {name, width, isFlexible};
+    }
```
cliffmeyers: I had a little trouble following what `name` is actually used for in the code, unless it's just there for debugging the list of columns? If it's truly optional, perhaps putting `isFlexible` as the second param would be a small improvement, since I can see the width and flexibility being something folks will use frequently.

src/js/components/jtable/TableRow.jsx
```diff
@@ -0,0 +1,145 @@
...
+                colWidth += TABLE_LEFT_RIGHT_PADDING + (TABLE_COLUMN_SPACING / 2);
+            } else {
+                colWidth += TABLE_COLUMN_SPACING;
+            }
+            newStyle.flexBasis = colWidth;
+
+            // Add or remove space on flexible columns in proportion to comparitive widths
+            newStyle.flexGrow = newStyle.flexShrink = columnDescription.isFlexible ? colWidth : 0;
+
+            return React.cloneElement(child, {style: newStyle});
```
cliffmeyers: This is a pretty elegant solution. 👍

cliffmeyers:
@michaelneale @i386 the second to last row doesn't have an href, so it's not linked anywhere. Suppose it's debatable whether a hover highlight should show in that case.

i386:
Perhaps should have a hover and the not allowed cursor ?

src/js/components/jtable/TableCell.jsx
sophistifunk: Keeps flow happy, because it's a const. Mutating things marked const is fucking stupid, but eslint insists on it. Because reasons.

src/js/components/jtable/JTable.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+    componentWillReceiveProps(nextProps: Props) {
+        if (nextProps.columns !== this.props.columns) {
+            this.processColumns(nextProps.columns);
+        }
+    }
+
+    processColumns(propColumns: any) {
+
+        if (propColumns == null) {
```
sophistifunk: That's why it's `==` instead of `===`

src/js/components/jtable/JTable.jsx
sophistifunk: It's used by `<TableHeaderRow/>` to generate header text, for convenience when you don't need arbitrary content in header cells. Not sure about it, I might re-evaluate it once I've had some more time applying this stuff to the main app.

sophistifunk:
That row in the "manual" example has no href on purpose to demonstrate link vs no-link rows. I might try and think of a better example to show that.

src/js/components/jtable/TableRow.jsx
sophistifunk: They serve different but related functions. You'll likely use `onClick` to do some sort of internal routing / ajaxy goodness, and the `href` functionality exists to allow middle-clicking into a new tab / window.

If the destinations aren't the same, we'll be hurting the user experience, but they definitely need to be separate mechanisms.

less/components/jtable.less
```diff
@@ -0,0 +1,96 @@
...
+//--------------------------------------
+
+@table-left-right-padding: 24px;
+@table-column-spacing: 32px;
+// !!!IMPORTANT!!! Don't change the above without changing the constants in JTable.jsx to match!
+
+// FIXME: We need to sort out the mess of different colours we've got all over the place
+@table-border-color: #e6eaee;
+@table-header-color: rgba(53, 64, 82, 0.5);
+@table-hover-bg: #fafbfd;
```
michaelneale: @cliffmeyers to track the pallette you mean? ie @brody may have something on that.

less/components/jtable.less
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+//--------------------------------------
+//  Cell Components
+//--------------------------------------
+
+.JTable-cell {
+
+    min-width: 0;
```
sophistifunk: Good catch. I'm not sure I quite remember why this is in there, so I'll pull it out and see what breaks ;-)

michaelneale:
right so no row highlight is by design. 

like @cliffmeyers said - would be nice to verify in IE11/edge at this point.

less/components/jtable.less
sophistifunk: https://issues.jenkins-ci.org/browse/JENKINS-42385

less/components/jtable.less
sophistifunk: Now I remember... 
screen shot 2017-03-01 at 11 33 20 am

sophistifunk: Added a comment in 5610451

sophistifunk:
Improved the examples to demonstrate the link / no link behaviour, as well as the ability to turn off rollover highlighting, or to turn it on for non-href rows.

i386:
@sophistifunk I had a problem trying to make the second example work in IE11. Seems to load fine in Chrome on my mac (same instance of story book - using it over the network from my windows vm.

ie11 - win7 running 2017-03-01 13-52-43

sophistifunk:
That's... Odd. We don't use Symbol, and I can't find a reference to it in our codebase. I will look into this further.

i386:
🎉  Works well on IE11

i386:
ie11 - win7 running 2017-03-01 17-36-03

src/js/components/jtable/JTable.jsx
cliffmeyers: Dur, yes, of course :)

src/js/components/jtable/TableCell.jsx
cliffmeyers: 👍

src/js/components/jtable/TableRow.jsx
cliffmeyers: I'm gonna go a little off-topic but bear with me :)

Generally React Router's `Link` component handles this behavior natively. We definitely don't want to couple the two, but as long as we can embed Link in a TableRow and the right thing happens, we're good to go.

Would the recommended usage pattern for users of React Router who want linked rows look like this?

```
<JTable columns={cols}>
   {rows}
</JTable>
```

Where a row is something like

```
<TableRow>
   <Link to="/some/url/somewhere">
       <TableCell>{person.name}</TableCell>
       <TableCell>{person.dob}</TableCell>
       <TableCell>{person.bloodType}</TableCell>
   </Link>
</TableRow>
```

I think as long as that use case is well supported, then we don't need to do anything with regard to `href` or `onClick`

less/components/jtable.less
brody: @michaelneale @cliffmeyers I'm trying to introduce some blue tint greys to reduce the harshness of the straight greys, with the intent to replace the current greys.

We/I need to find time to dedicate to solidifying style guide... whilst trying to iterate and improve upon what we currently have.

PR #157, opened 2017-03-13
Bug/jenkins 42719 dropdown fixes

cliffmeyers:
**Description**
- See [JENKINS-42719](https://issues.jenkins-ci.org/browse/JENKINS-42719).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
cc @sophistifunk

src/js/components/forms/Dropdown.jsx
```diff
@@ -56,6 +56,18 @@ export class Dropdown extends React.Component {
         return this.state.selectedOption;
     }
 
+    selectOption(option) {
```
sophistifunk: Just interested why a method that requires a ref rather than a prop? I'm worried we might be fighting against the "react way" a bit too much when we do stuff like this. Maybe we should separate out the job of maintaining "selected item" state into (the|a) container component..... Aaaaand now I'm rambling...

cliffmeyers: Yeah, good topic for discussion here. I would like to use props wherever possible, although the "interactive components" have a little wrinkle. I want to be careful to avoid forcing users of the components to have boilerplate code if it's possible. Using `Dropdown` as an example..

Dropdown has a `defaultOption` prop and a `selectedOption` state var. We need need to support three use cases:
1. Allow the Dropdown's "selected option" to be set during initial render. This is currently done with `defaultOption` prop.
2. Make sure the Dropdown's selected option is updated when the user picks an item in the list. This is done by updating the `selectedOption` state var when the user changes the item.
3. Allow the selected option to be to set programmatically at any time. This was going to be done via that `selectOption` method.

If the developer doesn't need to worry about 1 or 3, then it seems nice that the Dropdown tracks the item using its own internal state, so there isn't a need for tracking the selected option in the parent component's state (or elsewhere).

But maybe it makes more sense just to create a `selectedOption` prop on Dropdown. There seems to be a widely-accepted rule in the React world that component never update their own props. So for use case 2, it seems it the developer would need to track selectedOption in the state of the parent component, and update it via the change handler... which seems tedious, but maybe that makes the most sense.

Sorry for the length... does that make sense? Let me know if you have an inclination either way.

sophistifunk: Yeah I get that, and was thinking the same thing, but then I noticed that all the containing components had to keep track of the selected item anyway. When the parent receives the selection change event, it's storing that somewhere - it doesn't wait, and then ask the component what the current selection is at some future point to post off a request or something, which is what lead me to thinking about this.

cliffmeyers: Okay, we're in agreement then :) I guess it's just something we need to document as a pitfall and folks will get used to it.

This change would have an impact on other usages of Dropdown though. I know the parameterized run stuff is using it, and it might be in the editor as well. Perhaps @scherler and @kzantow could chime in here? I definitely don't want to regress any other parts of the UI this late in the game.

sophistifunk: We could extract the current behaviour into a fairly trivial wrapper component which just provides the state, exported but marked for future deletion, and update existing client code to use that instead, and mark it with TODOs to switch to the new one?

Too many commas :D

cliffmeyers: That's a good idea, definitely using composition would be great.

I wonder if we should implement the wrappers with the existing names, and possibly come up with new names for the components (not necessarily easy) or export them off a new topic level object, eg

```
import { Controls } from '@jenkins-cd/design-lanaguage`;
const { Checkbox } = Controls;
```

sophistifunk: I had that originally (the `components` namespace object), and everybody bitched about it constantly, so I'll say no. Apparently it's better to have one giant noisy namespace for everything.

cliffmeyers: So just got back to this one. I don't think now is the best time to begin these larger changes for two reasons:
- They will commit us to a path that maybe not everyone is onboard with: @scherler had some objections to namespacing in the Gitter room yesterday.
- It will change behavior in Dropdown and I don't want to introduce risk of regression into parameterized runs and the editor.

I'm going to remove `selectOption` from this PR so we don't expose a method that we plan to remove later. I can accomplish the same behavior via a single LOC inside of Git creation, document the hack, and then fix it once after we get everyone on the same page and do the implementation in a reasonable time frame.

cliffmeyers: @sophistifunk if you're curious about where I was originally going with the "composed component" approach you can view commit d1bea54 which I later reverted.

sophistifunk:
I'm confused, howcome we have two different `Dropdown` components and files now? I *really* don't like having multiple components or files with the same name when we're stuck with code tools that reply on grep and filename searches for navigation.

cliffmeyers:
@sophistifunk apologies, there was one commit I had forgot to revert. Should be clean now, mostly superficial changes in this PR now.

sophistifunk:
That makes a lot more sense 😁 

bb-tf

PR #159, opened 2017-03-22
Feature/jenkins 41740 placeholder

cliffmeyers:
**Description**
- Create a `PlaceholderTable` and related components for building the background in the attached ticket. Every screen has a subtly different background, so one generic one was not enough.
- See [JENKINS-41740](https://issues.jenkins-ci.org/browse/JENKINS-41740).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
cc @sophistifunk lord of the tables.

sophistifunk:
Aw man, how did I end up lord of the tables. I was much happier just being Baron of Brisbane (ITM).

src/js/components/Placeholder.jsx
```diff
@@ -0,0 +1,126 @@
+import React, { PropTypes } from 'react';
```
sophistifunk: Can we get a `//@flow`?

src/js/stories/PlaceholderStories.jsx
```diff
@@ -0,0 +1,89 @@
...
+        { width: 125, head: { text: 50 }, cell: { text: 100 } },
+        { width: 125, head: { text: 50 }, cell: { text: 100 } },
+        { width: 50, head: { text: 50 }, cell: { icon: 20 } },
+    ];
+
+    return (
+        <div style={style}>
+            <p>PlaceholderTable driven by 'columns' prop</p>
+
+            <PlaceholderTable columns={columns} style={{maxHeight: 415}} />
```
sophistifunk: Not sure what the `maxHeight` is for? Is it a leftover from dev / testing? I kind of like the idea of having the placeholder cut off, but a fade would be a lot nicer that a chop mid-row.

sophistifunk:
Couple of small comments, but the results looks great in Storybooks! 🐝

i386:
Sir Josh McDonald, Lord Of Tables, Protector of the Bezier's, Champion of the Stogie and Baron of Brisbane.

src/js/components/Placeholder.jsx
cliffmeyers: I added flow. I had to make two small tweaks to the Table components to get them to play nice:
- 0f3b5c9, made "children" optional since using `JTable` or `TableRow` in declarative fashion (within `Placeholder`) yielded errors. Googling this, it appears that the way Flow handles React children is pretty shit and there's a better workaround, unless you're aware of one?
- 2e24933, made column def's `name` optional since `processColumns` was already treating it as such.

src/js/stories/PlaceholderStories.jsx
cliffmeyers: I removed this. Originally I had set up the `Placeholder` to hide overflow of the inner `JTable` and was testing that, but it probably makes more sense to let the user of `Placeholder` decide how to handle it.

sophistifunk:
bbm-bueno

PR #165, opened 2017-05-08
Bug/use link for tablerow

sophistifunk:
**Description**

Changes `<TableRow>` to add the ability to specify `linkTo="foo"` instead of `href="foo"` in which case it will make the row element using react-router's `Link` component instead of just `<div>` or `<a>`.

Suggested by @cliffmeyers ages ago, I am just a lazy jerk who kept putting it off until it broke some stuff 😁

### DIFF
--- a/src/js/components/jtable/TableRow.jsx
+++ b/src/js/components/jtable/TableRow.jsx
@@ -1,7 +1,8 @@
 // @flow
 
 import React, { Component, PropTypes, Children } from 'react';
-import { TableHeader } from './';
+import { Link } from 'react-router';
+import { TableHeader } from '../';
 import {
     TABLE_COLUMN_SPACING,
     TABLE_LEFT_RIGHT_PADDING
@@ -82,6 +83,7 @@ export class TableRow extends Component {
             children,
             columns,
             href,
+            linkTo,
             onClick,
         } = this.props;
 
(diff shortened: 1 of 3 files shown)


michaelneale:
:bee: from me FWIW

src/js/components/jtable/TableRow.jsx
```diff
@@ -13,6 +14,7 @@ type Props = {
     className?: string,
     children?: ReactChildren,
     href?: string,
+    linkTo?: string,
```
cliffmeyers: @sophistifunk I like the functionality but I am a little worried about coupling the JDL to `react-router`.

What if we introduced a `link` prop of `PropTypes.element` that allowed the client code to pass in a React element used to wrap the row? We should just be able to `React.clone` the `link` and then pass the `TableRow` as its children?

sophistifunk: I don't want anything wrapping the rows and cells, so the Link has to be the row. And JDL is already dependent on react-router for other stuff.

cliffmeyers: You're right, it doesn't need to wrap, but it can just use the passed-in element in place. I may be wrong, but I don't see `react-router` as a dep in package.json (on master anyways).

sophistifunk: It's a devDependency, but it probably should be peer. But `<TabLink/>` already uses it, and I don't want to push that complexity into n instances of client code, when it can be in this library. IMO That's what libraries are for.

sophistifunk:
I'm gonna merge this, for the sake of progress. If we need to make it more flexible in the future we can.

PR #166, opened 2017-05-08
JENKINS-39635: Supress rightclick on tests

imeredith:
**Description**

https://github.com/jenkinsci/blueocean-plugin/pull/984

- See [JENKINS-39635.](https://issues.jenkins-ci.org/browse/JENKINS-XXXXX).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


src/js/components/ResultItem.jsx
```diff
@@ -58,7 +58,7 @@ export class ResultItem extends Component {
     }
 
     toggleExpanded: Function = (e) => {
-        if (this.props.children) {
+        if (this.props.children && !getSelection().toString()) {
```
cliffmeyers: `getSelection` might be undefined if we ever unit test this component in headless environment; you might want to do a simple check for it. Not critical though.

michaelneale:
```

 61:         if (this.props.children && !getSelection().toString()) {
                                         ^^^^^^^^^^^^^^ function call
 61:         if (this.props.children && !getSelection().toString()) {
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `toString`. Method cannot be called on possibly null value
 61:         if (this.props.children && !getSelection().toString()) {
                                         ^^^^^^^^^^^^^^ null

```

src/js/components/ResultItem.jsx
michaelneale: it was picked up by flow or something as an error so - yes.

cliffmeyers:
@imeredith If I were to select some text from a log that was expanded, and then clicked on the header to collapse it, would this code prevent that from happening? If yes, then we may need to take a different approach. @i386 WDYT?

i386:
@cliffmeyers @imeredith if indeed that is the case then we need a different approach.

cliffmeyers:
@i386 I tested this quickly: the issue is not as severe as I thought it might be. If you select text anywhere outside of the `ResultItem` header, then click the header, it will first deselect the text before triggering the click, so the `ResultItem` will toggle open/close as you expect. However, if you've selected the text in a `ResultItem` header, then click it, the first click will only clear the selection, not toggle the header. You need to click a second time to trigger the toggle.

PR #167, opened 2017-05-10
CommitHash component should pass through invalid values

i386:
**Description**

Instead of pushing Git specific code into a extension point (which are going to change in a huge way soon) for the time being we can just pass through the "invalid" commit hashes. This means people using subversion, perforce and mercurial should be able to see their commit identifiers wherever we put `CommitHash`

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


i386:
@cliffmeyers thanks for the review. I think it might be worth doing that when we tackle making this pluggable (there is a ticket tracked in JIRA).

i386:
@cliffmeyers I went ahead and did what you asked anyway - was easy 👍

PR #170, opened 2017-05-19
Experiment/form experiments

cliffmeyers:
**Description**
*DO NOT MERGE* - just for illustrate purposes and discussion.

This is an example of three different approaches to form controls:
- "Classic" is analogous to how the current controls work. They manage their own state as the user interacts but there are some bugs and non-ideal woarkarounds to how their state can be managed programmatically from a parent.
- "Stateless" is probably the best option, which requires a little more boilerplate so it will update as user interacts, but in practice we might need that code anyways. This seems to be the most flexible and not suffer from any bugs that I've seen yet.
- "Hybrid" was an attempt to meld the two together but it seems to suffer from a bug where you can't reliably reset the control to its default value after the user has typed.

### DIFF
--- a/src/js/stories/form-experiment/TextInputClassic.jsx
+++ b/src/js/stories/form-experiment/TextInputClassic.jsx
@@ -1,0 +1,108 @@
+import React, { Component, PropTypes } from 'react';
+
+
+export default class TextInputClassicStory extends Component {
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            defaultValue: 'default value passed before mounting',
+        };
+
+        this.textRef = null;
+    }
+
+    _setDefaultValue(value) {
+        console.log('_setDefaultValue:', value);
+        // this doesn't work since defaultValue is only read from when TextInputClassic mounts
+        this.setState({
+            defaultValue: value,
+        });
+    }
+
+    _setValue(value) {
+        console.log('_setValue:', value);
+        this.textRef.setValue('setValue: ' + value);
+    }
+
+    render() {
+        return (
+            <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-start', height: 300, justifyContent: 'space-between', padding: 10 }}>
+                <h1>Classic</h1>
+
+                <span>
+                    <TextInputClassic />
+                    &nbsp; updates as you type without extra code, without an initial value set
+                </span>
+
+                <span>
+                    <TextInputClassic defaultValue={this.state.defaultValue} />
+                    &nbsp; updates as you type without extra code, with an initial value set
+                </span>
+
+                <br />
+
+                <span style={{ background: '#FF9999'}}>
+                    <button onClick={() => this._setDefaultValue(new Date().toString())}>Set defaultValue</button>
+                    &nbsp; Won't work, since defaultValue is only read by text control when mounting
+                </span>
+
+                <span>
+                    <button onClick={() => this._setValue(new Date().toString())}>Call setValue</button>
+                    &nbsp; Will work, but requires a ref, and is less "reactive"
+                </span>
+
+                <TextInputClassic ref={ref => this.textRef = ref} defaultValue={this.state.defaultValue} />
+            </div>
+        );
+    }
+}
+
+
+class TextInputClassic extends Component {
+
+    propTypes: {
+        defaultValue: PropTypes.string,
+    };
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            value: '',
+        };
+    }
+
+    componentWillMount() {
+        this._initialize(this.props);
+    }
+
+    _initialize(newProps) {
+        if (!this.state.value && newProps.defaultValue) {
+            this.setState({
+                value: newProps.defaultValue,
+            });
+        }
+    }
+
+    _onChange(event) {
+        this.setState({
+            value: event.currentTarget.value,
+        });
+    }
+
+    setValue(value) {
+        this.setState({
+            value,
+        });
+    }
+
+    render() {
+        return (
+            <input type="text" style={{ width: 400 }} value={this.state.value} onChange={e => this._onChange(e)} />
+        );
+    }
+
+
+}
--- a/src/js/stories/form-experiment/TextInputClassic.jsx
+++ b/src/js/stories/form-experiment/TextInputClassic.jsx
@@ -1,13 +1,16 @@
 import React, { Component, PropTypes } from 'react';
 
 
+const DEFAULT_VALUE = 'default value passed before mounting';
+
+
 export default class TextInputClassicStory extends Component {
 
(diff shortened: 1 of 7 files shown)


src/js/stories/form-experiment/TextInputClassic.jsx
```diff
@@ -0,0 +1,108 @@
...
+                    <TextInputClassic defaultValue={this.state.defaultValue} />
+                    &nbsp; updates as you type without extra code, with an initial value set
+                </span>
+
+                <br />
+
+                <span style={{ background: '#FF9999'}}>
+                    <button onClick={() => this._setDefaultValue(new Date().toString())}>Set defaultValue</button>
+                    &nbsp; Won't work, since defaultValue is only read by text control when mounting
+                </span>
```
cliffmeyers: This illustrates the bug that @NicuPascu found, as component won't apply `defaultValue` prop after it has mounted.

src/js/stories/form-experiment/TextInputClassic.jsx
```diff
@@ -0,0 +1,108 @@
...
+        super(props);
+
+        this.state = {
+            value: '',
+        };
+    }
+
+    componentWillMount() {
+        this._initialize(this.props);
+    }
```
cliffmeyers: We could try to fix the bug by adding `componentWillReceiveProps` and letting it apply the `defaultValue` prop, but it won't work because any prop change would trigger this, and the value the user typed would get reset.

cliffmeyers:
form-experiments-1-classic
form-experiments-2-stateless
form-experiments-3-hybrid

src/js/stories/form-experiment/TextInputHybrid.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+                <span>
+                    <button onClick={() => this._setValue('')}>Clear 'value' prop</button>
+                    &nbsp; Works fine, since parent has full control over text's state
+                </span>
+
+                <span style={{ background: '#FF9999'}}>
+                    <button onClick={() => this._setValue(DEFAULT_VALUE)}>Reset 'value' prop</button>
+                    &nbsp; Doesn't work after typing, since the value of prop 'value' didn't change
+                </span>
```
cliffmeyers: This illustrates the "unable to reset" bug, which I'm not sure can really be fixed in any reliable way. The workaround would be to clear "value" and then set it to the default value... pretty ugly.

src/js/stories/form-experiment/TextInputStateless.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+                <span>
+                    <TextInputStateless onChange={value => this._onChange(value)} />
+                    &nbsp; updates as you type without extra code, without an initial value set
+                </span>
+
+                <span style={{ background: '#FF9999'}}>
+                    <TextInputStateless value={this.state.value1} />
+                    &nbsp; <b>will not update</b> as you type without extra code, because an initial value set
+                </span>
```
cliffmeyers: This is only bug I found in the "stateless" component, although you just need a change handler defined that updates state in the parent component... which we'll probably do anyways. We could even log a warning in the component if no change handler is provided.

src/js/stories/form-experiment/TextInputClassic.jsx
```diff
@@ -0,0 +1,121 @@
...
+        this.setState({
+            value: event.currentTarget.value,
+        });
+    }
+
+    setValue(value) {
+        this.setState({
+            value,
+        });
+    }
```
cliffmeyers: The workaround would be to expose a method like this, but as you see above it requires a ref to the component, and we aren't quite embracing the React philosophy as much since we are updating imperatively rather than pushing down a prop.

src/js/stories/form-experiment/TextInputStateless.jsx
kzantow: ... or make it an error, required prop

src/js/stories/form-experiment/TextInputStateless.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+                <TextInputStateless onChange={value => this._setValue(value)} value={this.state.value2} />
+            </div>
+        );
+    }
+}
+
+
+class TextInputStateless extends Component {
```
sophistifunk: This is how we should be doing it going forward IMO. It matches the way react normally behaves when using text components. I definitely put a little too much state in some of our earlier components, which makes things difficult when you're trying to blend in with react's "props down, events up" model.

michaelneale:
nothing to add but nice catch @cliffmeyers and @NicuPascu !

(buidl failure is just a few lint complains about props)

src/js/stories/form-experiment/TextInputStateless.jsx
cliffmeyers: Agreed. When I stated these experiments the other day, my hope was that I could devise a "best of both worlds" solution but I ultimately found bugs in the "Hybrid" impl that are just too unpredictable.

cliffmeyers:
To resolve this, I think the following steps need to be taken:
- Build new components that use the "stateless" approach. This should be a pretty simple copy/paste of the old components that exposes a `value` or `selectedValue` prop, eliminates the `defaultValue` prop and makes a `onChange` handler prop required.
- Figure out where to place the new comps and how to export them so they can live side-by-side the old components.
- Deprecate the old components; consider logging a warning the first time a component mounts to inform developer of the deprecation.
- Add some JIRAs to migrate over the components. I don't think there's a huge rush to do it; most importantly we just need to ensure new development uses the new component.

I know that sounds like a lot of work, but I don't think we can just blow up the universe by changing the existing components.

i386:
@cliffmeyers if there is work here would you mind ticketing it up? We can find a time to make those improvements cc @michaelneale

src/js/stories/form-experiment/TextInputClassic.jsx
scherler: @cliffmeyers 
> We could try to fix the bug by adding componentWillReceiveProps and letting it apply the defaultValue prop, but it won't work because any prop change would trigger this

well you could add an evaluation like `nextProps.deauftValue !== this.state.defaultValue` to `componentWillReceiveProps` or maybe even `shouldComponentUpdate`

cliffmeyers: I thought the same thing but it won't work in this scenario :/

- input has `defaultValue` prop set
- user begins to type in the text field (updating `state.value`)
- some other prop on the input changes
- Logic will see that `props.defaultValue !== state.value` and clobber what the user just typed

cliffmeyers:
Summarized work to be done in:  https://issues.jenkins-ci.org/browse/JENKINS-44622

Closing this PR as there is no intent to merge it.

PR #171, opened 2017-05-24
Task/arbitrary table props

sophistifunk:
**Description**

Forward arbitrary props in table components, mainly to support `data-X` attribs, but also for `title` etc.

@reviewbybees

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@jenkins-cd/design-language",
   "jdlName": "jenkins-design-language",
-  "version": "0.0.130-SNAPSHOT",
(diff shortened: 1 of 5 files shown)


sophistifunk:
*ping* @cliffmeyers @kzantow if you've got a minute.

cliffmeyers:
I'm sure this change is probably fine, could you just explain how `restProps` is getting defined? Is using `... someVar` at the end of the destructuring assignment taking all the other props from `this.props` except those explicitly defined beforehand?

sophistifunk:
Yeah, it's super-useful for this sort of thing.

michaelneale:
:bee: I guess. data tags are most useful.

PR #173, opened 2017-06-05
Task/export truncating label

cliffmeyers:
**Description**
- Just exporting TruncatingLabel so it can be used in the app

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@jenkins-cd/design-language",
   "jdlName": "jenkins-design-language",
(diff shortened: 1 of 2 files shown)


michaelneale:
this will need to be moved to the main blueocean-plugin repo @cliffmeyers

cliffmeyers:
I'll just close this one. If we need truncating label elsewhere we can export in a separate PR.

PR #174, opened 2017-06-16
Bug/broken actions column

sophistifunk:
**Description**

* Various small tweaks to styles
* Adds the ability to specify links on cells as well as on rows
* Some storybook updates

@reviewbybees


src/js/components/jtable/TableRow.jsx
```diff
@@ -21,6 +21,35 @@ type Props = {
...
+ * Generate the props and optionally override the element tag / component required to implement a link on a row or cell.
+ */
+export function generateLink(defaultTagOrComponent: TagOrComponent, href ?: string, linkTo ?: string) {
+    let isLink = false;
+    let linkProps = undefined;
+    let tagOrComponent: TagOrComponent = defaultTagOrComponent;
+
+    if (typeof href === 'string' && href.length > 0) {
+        isLink = true;
+        // We switch to an <A> instead of <DIV> so the user can middle-click
```
kzantow: middle-click?? pfff

less/components/jtable.less
```diff
@@ -81,6 +81,11 @@ a.JTable-row--href {
     padding-left: @table-column-spacing / 2;
     padding-right: @table-column-spacing / 2;
 
+    a&, a&:hover {
```
kzantow: I had to look this syntax up. *sigh*

src/js/components/jtable/TableCell.jsx
```diff
@@ -24,34 +37,40 @@ export class TableCell extends Component {
             classNames.push(className);
         }
 
+        const {
+            linkProps,
+            tagOrComponent
+        } = generateLink('div', href, linkTo);
```
kzantow: so the whole point of this is to use things other than `div` as the wrapping element, right?

less/components/jtable.less
sophistifunk: It's stupid having to do these tricks to increase specificity. Almost as stupid as having to do all this work because you can't wrap a `<tr>` with `<a>`

src/js/components/jtable/TableCell.jsx
kzantow: why do we even have `TableCell`?? why not just build in logic to `TableRow` or styles based on immediate child components? it would be more flexible that way and less code for the end user...

sophistifunk: Sometimes

kzantow: I guess, I find it really odd that we're using a `TableCell` as a random "do-everything" container, that's not an approach that I like at all. I wrote a bunch of other stuff, but it was irrelevant, so I'll leave it at that.

sophistifunk: It's not a random do everything container, it's a stand-in for `<td>` so that when something else comes up that's a problem in the future we can change how tables are generated without having to refactor every single component that uses tables.

sophistifunk:
...None of which applies to this change that simply duplicates the link generation ability that existed on whole-rows to single-cells, and a handful of lines of CSS that used to be in -web but belong here.

PR #176, opened 2017-06-20
Bug/jenkins 42131 jdl dropdowns in wrong place in firefox

NicuPascu:
**Description**
- See [JENKINS-42131](https://issues.jenkins-ci.org/browse/JENKINS-42131).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/less/components/popups.less
+++ b/less/components/popups.less
@@ -9,7 +9,7 @@
 //--------------------------------------
 
 .popup-shadow {
-    filter: drop-shadow(0px 2px 4px rgba(0, 0, 0, 0.5));
+    box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.5);
 }
 
 @popup-border-radius: 3px;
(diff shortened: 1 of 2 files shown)


less/components/popups.less
```diff
@@ -9,7 +9,7 @@
 //--------------------------------------
 
 .popup-shadow {
-    filter: drop-shadow(0px 2px 4px rgba(0, 0, 0, 0.5));
+    box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.5);
```
sophistifunk: Aw man, this is all we had to do to fix this!?!? I thought it was gonna be a huge shitfight!

sophistifunk:
Nice work!

i386:
Bravo!!

less/components/popups.less
NicuPascu: yup, this happens in ff because of this: https://bugzilla.mozilla.org/show_bug.cgi?id=876341

makes any container that has a "transform" effect applied to if act as a container for fixed positioned elements instead of the viewport

cliffmeyers: Wow, that's nasty, but awesome you found such a simple fix!

PR #178, opened 2017-06-27
bug/npm-publishing-fail 

sophistifunk:
* Work-around for a 'fix' somewhere in npm that makes `.gitignore` override `files` array in package.json


michaelneale:
:bee: so it's nice and legal

PR #180, opened 2017-07-03
Bug/jenkins 45011 commit in run result header is not clickable

NicuPascu:
**Description**
re-factor the CommitId component so it can accept an url and generate a link in case it is present
- See [JENKINS-45011](https://issues.jenkins-ci.org/browse/JENKINS-45011).

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/src/js/components/CommitId.jsx
+++ b/src/js/components/CommitId.jsx
@@ -9,7 +9,7 @@ const GIT_HASH_REGEX = /\b[0-9a-f]{5,40}\b/;
  */
 export class CommitId extends Component {
     render() {
-        const {commitId, url} = this.props;
+        const {commitId, url, title} = this.props;
         let displayValue;
         if (GIT_HASH_REGEX.test(commitId)) {
             displayValue = commitId.substring(0, 7);
@@ -20,7 +20,7 @@ export class CommitId extends Component {
         }
 
         if (url) {
-            return (<a href={url} target="_blank" title="Opens commit in a new window">
+            return (<a href={url} target="_blank" title={title ? title : 'Opens commit in a new window'}>
                 <code className="hash">{displayValue}</code>
             </a>);
         }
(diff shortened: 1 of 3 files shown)


src/js/components/CommitId.jsx
```diff
@@ -18,10 +18,18 @@ export class CommitId extends Component {
         } else {
             displayValue = '–';
         }
+
+        if (url) {
+            return (<a href={url} target="_blank" title="Opens commit in a new window">
```
cliffmeyers: I would suggest adding an optional "title" prop that defaults to this string value. We want to make sure we can apply i18n here.

test/js/components/CommitId-spec.jsx
```diff
@@ -24,4 +24,10 @@ describe("CommitId", () => {
         assert.equal(wrapper.text().length, 7);
         assert.equal(wrapper.text(), '676b757');
     });
+
+    it("renders link when url prop is passed along", () => {
+        const wrapper = shallow(<CommitId commitId="123" url="link" />);
+        assert.isTrue(wrapper.is('a'));
+        assert.equal(wrapper.html(), '<a href="link" target="_blank" title="Opens commit in a new window"><code class="hash">123</code></a>');
```
cliffmeyers: You could make a change here too but I don't think it's absolutely required. Generally with Enzyme I use one of two different approaches:
- Test for presence of nodes with `find()` or against node values using `prop(name)`, `text()` and similar. This seems to be the most popular thing to do but can be tedious if you have a lot of nested nodes - especially nested components - since you need to call `shallow()` on each comp. (Although at that point, it's more of an integration test than a pure unit test).
- Test with `html()` and `indexOf` to check that HTML contains certain string fragments you might expect. This tends to be very fast and easy to work with.

I would be cautious about using `html()` and an exact string comparison just because it will break in the event of any change made to the component.

src/js/components/CommitId.jsx
i386: Good idea!

michaelneale:
@NicuPascu this will need to be moved to the dir in the blueocean-plugin repo

PR #181, opened 2017-07-03
Update Jenkinsfile

michaelneale:
**Description**
 
move to blue ocean repo


michaelneale:
ok @kzantow this should serve as a notice

kzantow:
Nice one!

PR #200, opened 2018-07-23
Sidebar collapse updates

lausena:
Sidebar collapse handling update

### DIFF
--- a/components/Sidebar/Sidebar.tsx
+++ b/components/Sidebar/Sidebar.tsx
@@ -16,18 +16,21 @@ interface SidebarProps {
     main: React.ReactNode;
     sheet?: React.ReactNode & { props: SidebarSheetProps };
     footer?: React.ReactNode;
-    collapse: boolean;
+    collapse?: boolean;
 }
 
 export class Sidebar extends React.Component<SidebarProps> {
+    public static defaultProps: Partial<SidebarProps> = {
+        collapse: false,
+    };
+
     static AppLogo = AppLogo;
     static Item = SidebarItem;
 
     nav: SidebarBehavior;
 
     componentDidMount() {
-        const { collapse } = this.props;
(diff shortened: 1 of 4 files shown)


components/Sidebar/Sidebar.tsx
```diff
@@ -16,16 +16,21 @@ interface SidebarProps {
...
+    };
+
     static AppLogo = AppLogo;
     static Item = SidebarItem;
 
     nav: SidebarBehavior;
 
     componentDidMount() {
-        this.nav = new SidebarBehavior();
+        this.nav = new SidebarBehavior(this.props.collapse || false);
```
mkramb: Just a suggestion, in the constructor we could do:

```
constructor(props: SidebarProps) {
    super(Object.assign({}, Sidebar.defaultsProps, props);
}
```

So that we rely on `defaultProps` definitions.
We could use that as a commn pattern.

components/Sidebar/SidebarBehavior.ts
```diff
@@ -32,14 +32,15 @@ export class SidebarBehavior {
...
+    constructor(collapse: boolean, onResize?: NavigationResized) {
         // Needed elements:
         this.body = $('body');
         this.nav = $('.Nav');
         this.navHandle = $('.NavExpander');
+        this.isCollapsed = collapse;
 
         // Interaction variables
-        this.isCollapsed = 'true' === getStored('jdl.navbar.collapsed');
+        this.isCollapsed = collapse;
```
mkramb: I think it should be:

```
this.isCollapsed = collapse !== undefined ? collapse : 'true' === getStored('jdl.navbar.collapsed');
```
You don't want to overrite (from the prop) if the user has already interacted with the sidebar (was collapsed or expanded), we should probably rename  `collapse` to `collapsedByDefault`.

